PR #2, opened 2022-01-02
Exams

LikerFeed:
Ivan Anenko IM-13


tshemsedinov (commented):
- Use eslint and prettier
- Follow style and naming conventions
- Optimize lexical scope, minimize area of identifier visibility
- Use round brackets even for single argumanet lambda functions

tshemsedinov:
Why do you have 2 github accounts?

PR #4, opened 2022-01-02
Exams

chyrkovmaksym:


tshemsedinov (commented):
- Prefer arrow functions
- Use eslint and prettier
- Use round brackets even for single argumanet lambda functions
- Use intermediate variables, decompose long expressions

chyrkovmaksym:
Got it! Thanks!

пн, 3 янв. 2022 г., 13:40 Timur Shemsedinov ***@***.***>:

PR #5, opened 2022-01-02
Exams

volkovily:


tshemsedinov (commented):
- Prefer arrow functions
- Remove unneeded operations, calls, arguments, blocks, etc.
- Add let or const for all identifiers, prefer const
- Decompose complex functions to single responsibility
- Use separation of concerns principle
- Respect 2 space indentation
- Use eslint and prettier

volkovily:
Used eslint and prettier

PR #6, opened 2022-01-02
Exams

just-hoshinon:


tshemsedinov (commented):
- Prefer arrow functions
- Use eslint and prettier

Tasks/plane.js
```diff
@@ -1,19 +1,19 @@
...
-    arr[i] = res[i - 1]
-    }else{
-    arr[i] = res[j - 1]
-    res.push(value);
-    };
-  };
-  return res
-}
+const planeArray = (arr) => {
+  const res = []
```
tshemsedinov, line 5:
```suggestion
  const res = [];
```

Tasks/take.js
```diff
@@ -1,11 +1,15 @@
...
 
-tAKe=(DX,       ...xor)=>{
-T = Object.keys(DX);
-T.forEach((_) => {
-() => 5;
-if (xor.includes(_)){} else delete DX[_];},21);
-return 'a', 'b', 'c', 'd', DX};
+const take = (object, ...keys) => {
+  const result = {};
+  for(const [key, value] of Object.entries(object)){
```
tshemsedinov, line 6: Use eslint, add spaces

Tasks/take.js
```diff
@@ -1,11 +1,15 @@
...
+    if(keys.includes(key)){
+      result[key] = value;
+    }
+  }
+  return result;
+};
 
-const result = tAKe({ a: 'uno', b: 'due', c: 'tre' }, 'b', 'c');
+const result = take({ a: "uno", b: "due", c: "tre" }, "b", "c");
```
tshemsedinov, line 14: I'd prefer single quotes

PR #7, opened 2022-01-02
Exams

AidXylelele:


Tasks/count.js
```diff
@@ -1,12 +1,16 @@
...
+'use strict';
 
-count =obj=> {sum=0;{}
- keys = Object.keys(obj);
-   keys.forEach((key) => {
-      ([]);const value = obj[key];
-      if (typeof value === 'number')sum += value });
-     ({});
-   return sum};
+const count = obj => {
```
tshemsedinov, line 4: Use round brackets even for single argumanet lambda functions

Tasks/count.js
```diff
@@ -1,12 +1,16 @@
...
+  let sum = 0;
+  for (const key in obj) {
```
tshemsedinov, line 6: Prefer for..of here

Tasks/count.js
```diff
@@ -1,12 +1,16 @@
...
+    const value = obj[key];
+    if (typeof value === 'number') {
+      sum += value;
+    }
+  }
+  return sum;
+};
 
-const result = count({ a: 1, b: 'two', c: 3, d: 4 });
+const result = count({ a: '', b: 'two', c: '', d: '' });
```
tshemsedinov, line 15: Don't change this line, it's like a small unit-test

Tasks/intersection.js
```diff
@@ -1,17 +1,14 @@
...
-}} else                                                {
-delete object_1[attribute_name];                       {
-                                                       }
-                  }{}
-                  }{}
-return object_1;
-}
+const intersection = (obj1, obj2) => {
+  const res = {};
+  for (const key in obj1) {
```
tshemsedinov, line 5: Prefer for..of here

PR #13, opened 2022-01-02
Exams

AndrijShchegel:
Щегель Андрій ІМ-11

### DIFF
--- a/Tasks/ip.js
+++ b/Tasks/ip.js
@@ -1,4 +1,5 @@
 // Split string by the first occurrence of separator
+'use strict';
 
 Parseip = i => {
   a = []
--- a/Tasks/ip.js
+++ b/Tasks/ip.js
@@ -2,7 +2,7 @@
 'use strict';
 
 const parseIp = ip => {
-  const res = []
+  const res = [];
   if (ip === '') return 0;
   else {
     const array = ip.split('.');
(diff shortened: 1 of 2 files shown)


Tasks/ip.js
```diff
@@ -1,17 +1,18 @@
 // Split string by the first occurrence of separator
+'use strict';
 
-Parseip = i => {
-  a = []
-    if (i === '' ) return 0  ;
+const parseIp = (ip) => {
+  const res = [];
+  if (ip === '') return 0;
   else {
```
tshemsedinov, line 7: You do not need else at all

Tasks/skip.js
```diff
@@ -1,17 +1,18 @@
 // Remove values from array
+'use strict';
 
-Skip =(T,...Remove)=> {
-x = 0;
-for (C of T) {
-for (X of Remove) {
-if (C === X) {
-  T.splice(x, 1);
+const skip = (array,...remove) => {
```
tshemsedinov, line 4: - Use eslint and prettier
- Follow style and naming conventions

Tasks/skip.js
```diff
@@ -1,17 +1,18 @@
...
+  let index = 0;
+  for (const value of array) {
+    for (const valueToDel of remove) {
```
tshemsedinov, line 7: We certainly do not need nested loops here

PR #14, opened 2022-01-02
Exams

AlexKirovich:
Кирович Александр ИМ-13


Tasks/between.js
```diff
@@ -1,18 +1,14 @@
...
-          return  '' } else { str =  str.substring(0 , i)
-            }}}
-  return str
-}
+const getValueBetween = (str, start, end) => {
+  const firstIndex = str.indexOf(start);
+  const lastIndex = str.indexOf(end);
+  let result = '';
+  if (firstIndex === -1 || lastIndex === -1) return '';
+  else result = str.substring(firstIndex + start.length, lastIndex);
```
tshemsedinov, line 9: We don't need else here

Tasks/intersection.js
```diff
@@ -1,17 +1,16 @@
...
-delete object_1[attribute_name];                       {
-                                                       }
-                  }{}
-                  }{}
-return object_1;
-}
+const intersection = (obj1, obj2) => {
+  const result = {};
+  const keys = Object.keys(obj1);
+  for (const key of keys){
```
tshemsedinov, line 7: Use eslint and prettier

PR #18, opened 2022-01-02
Exams

vladimirvikulin:


vladimirvikulin:
Perhaps, it is beter to use for...of here, but I`m not sure

tshemsedinov:
For..of sure

PR #19, opened 2022-01-02
Exams

GhostDolphin:


Tasks/ip.js
```diff
@@ -1,17 +1,3 @@
...
-    for ( const b of B) {
-  a[j] = parseInt(b);
-  j++;
-  }};
-  return a;
-};
-
-const result = Parseip('127.0.0.1')
-console.log(result)
+const parseIp = ip => ip === '' ? 0 : ip.split('.').map(e => +e);
```
tshemsedinov, line 3: - Use parseInt is preferable here
- Intermediate variables will make this code more readable
- Use round brackets even for single arguments lambda functions
- Prefer if-statement here for readability
- Improve naming, `e` is not ok

Tasks/plane.js
```diff
@@ -1,19 +1,8 @@
...
-    arr[i] = res[j - 1]
-    res.push(value);
-    };
-  };
-  return res
-}
-
-const result = const_plane([1, [[2]], [3, 4], [5], [6, [7, 8]]]);
-console.log(result);
+const plane = arr => {
```
tshemsedinov, line 3: Use round brackets even for single argument lambda functions

Tasks/plane.js
```diff
@@ -1,19 +1,8 @@
...
+  const res = [];
+  for (const val of arr)
+    Array.isArray(val) ? res.push(...plane(val)) : res.push(val);
```
tshemsedinov, line 6: - Use if-statement instead of ternary here
- Add code block

Tasks/ip.js
GhostDolphin: ip.js is fully fixed

Tasks/plane.js
GhostDolphin: Fixed

Tasks/plane.js
GhostDolphin: plane.js is fully fixed

Tasks/ip.js
```diff
@@ -1,17 +1,9 @@
...
-  else {
-  B = i.split('.');
-  j = 0
-    for ( const b of B) {
-  a[j] = parseInt(b);
-  j++;
-  }};
-  return a;
+const parseIp = (ip) => {
+  if (ip !== '') {
```
tshemsedinov, line 4:
```suggestion
  if (ip === '') return null; // or undefined or null or even empty array []
```

PR #21, opened 2022-01-02
Exams

GAlexandrD:


GAlexandrD:
It seems to me that the closure option is more optimized because we are not creating temporary arrays, moreover local constant option uses a lot of unnecessary push calls and rest operators. So we can use a more suitable option for our purposes.

PR #23, opened 2022-01-03
Exams

vladyatsuk:


tshemsedinov (commented):
- Use eslint and prettier
- Respect 2 space indentation

PR #24, opened 2022-01-03
Exams

Nailgy:


Nailgy:
Note: there is a wrong naming of res in commit "Moving res from argument ...", res is a parameter of function

tshemsedinov:
just by the way: Such a caching is not required `const length = arr.length;` V8 engine will optimize it automatically

PR #26, opened 2022-01-03
Exams

Kirillbiliashov:


tshemsedinov (commented):
- Use eslint and prettier
- Respect 2 space indentation
- Prefer for..of

PR #27, opened 2022-01-03
Exams

vladmoroca:


tshemsedinov (commented):
- Use eslint and prettier and follow style and naming conventions
- Use intermediate variables

vladmoroca:
Where should I use intermediate variables?

PR #30, opened 2022-01-03
Exams

LLumisS:
Kirill Bukach
IM-11

### DIFF
--- a/Tasks/intersection.js
+++ b/Tasks/intersection.js
@@ -2,16 +2,16 @@
 
 // Finde an intersection of two dictionaries
 
-const intersection = (object_1, object_2) => {
-  first_keys = Object.keys(object_1);
-  for (attribute_name of first_keys) {
-    if (attribute_name !== '' && object_2[attribute_name]) {
-      object_2[attribute_name] = object_1[attribute_name];
+const intersection = (object1, object2) => {
+  const firstKeys = Object.keys(object1);
+  for (const attributeName of firstKeys) {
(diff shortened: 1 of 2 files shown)


Tasks/intersection.js
```diff
@@ -1,17 +1,17 @@
...
-                  }{}
-                  }{}
-return object_1;
-}
+const intersection = (object1, object2) => {
+  const result = {};
+  const keys = Object.keys(object1);
+  for (const key of keys) {
+    if (object1[key] === object2[key]) {
+      result[key] = object1[key];
```
tshemsedinov, line 10: Use intermediate variable

Tasks/replace.js
```diff
@@ -1,18 +1,22 @@
...
-  res = ''
-  do {
-   const _index = src.indexOf( substr) ; 
-   if (_index === -1) { return res + src }else{
-   const start = src.substring(0,_index);
-   src = src.substring(_index+substr.length,src.length)
-   res += start + newstr
- }} while (true); }; };
+const replace = (string, subString, newString) => {
+  let result;
```
tshemsedinov, line 6: Need initial value

Tasks/replace.js
```diff
@@ -1,18 +1,22 @@
...
+  const index = string.indexOf(subString);
+  if ((subString === '') || (index === -1)) {
+    result = string;
```
tshemsedinov, line 9: We can return here and do not need else block

PR #31, opened 2022-01-03
Exams

alinagts:


Tasks/words.js
```diff
@@ -1,26 +1,11 @@
...
-    if (c === ' ') {
-      if (flag === true) { flag = false } else { flag = false }
-    } else {
-      if (flag === true) { flag = true } else { flag = true }
-    }
-   }
- }
- return number_of_words_in_s
- }
+const words = string => {
```
tshemsedinov, line 4: Use round brackets even for single argument lambda functions

Tasks/words.js
```diff
@@ -1,26 +1,11 @@
...
+ const arrOfWords = string.split(" ");
+ let numberOfWords = arrOfWords.length;
```
tshemsedinov, line 6: Use const here

tshemsedinov (commented):
Use eslint and prettier and follow style

PR #38, opened 2022-01-03
Exams

getOutOfHereInSmallSteps:


Tasks/intersection.js
```diff
@@ -1,17 +1,14 @@
...
-                                                       }
-                  }{}
-                  }{}
-return object_1;
-}
+const intersection = (obj1, obj2) => {
+  const keys = Object.keys(obj1);
+  const res = {};
+  for (const key of keys) {
+    if (obj1[key] === obj2[key]) res[key] = obj1[key];
```
tshemsedinov, line 8: Long line, add block and/or use intermediate variables

getOutOfHereInSmallSteps: fixed

PR #39, opened 2022-01-03
Exam

AlinaDubchak:
Aded and optimized tasks (inc.js and ip.js)


Tasks/inc.js
```diff
@@ -1,14 +1,15 @@
-// Increment all numbers in dictionary
+'use strict'
```
tshemsedinov, line 1: - Use eslint and prettier

Tasks/inc.js
```diff
@@ -1,14 +1,15 @@
...
 
-let inc_numbers=(
```
tshemsedinov, line 3: - Prefer const

Tasks/inc.js
```diff
@@ -1,14 +1,15 @@
...
-  (format_complete,...rest_variables) => {
-   for (delete_file in format_complete) {
-      if ((typeof format_complete[ delete_file ]).charAt(0).toUpperCase() ==='N'){
-      format_complete[ delete_file ] = format_complete[delete_file] + 1}
+const inc = elements => {
```
tshemsedinov, line 3: Use round brackets even for single argument lambda functions

Tasks/inc.js
```diff
@@ -1,14 +1,15 @@
...
+     for ( let key in elements){
+      if ((typeof elements[key]) == "number")
```
tshemsedinov, line 5: Never use `==`

Tasks/inc.js
```diff
@@ -1,14 +1,15 @@
-// Increment all numbers in dictionary
+'use strict'
 
-let inc_numbers=(
-  (format_complete,...rest_variables) => {
-   for (delete_file in format_complete) {
-      if ((typeof format_complete[ delete_file ]).charAt(0).toUpperCase() ==='N'){
-      format_complete[ delete_file ] = format_complete[delete_file] + 1}
+const inc = elements => {
+     for ( let key in elements){
```
tshemsedinov, line 4: Prefer for..of

Tasks/inc.js
```diff
@@ -1,14 +1,15 @@
...
+      if ((typeof elements[key]) == "number")
+      {
+        elements[key] = elements[key] + 1;
```
tshemsedinov, line 7: ++

Tasks/inc.js
```diff
@@ -1,14 +1,15 @@
...
+      }
    }
-  return format_complete;
-  }
-)
+  return elements;
```
tshemsedinov, line 10: Don't change incoming parameters

tshemsedinov (commented):
Terrible formatting, plz fix all linter warnings

AlinaDubchak:
// another implementation using forEach.
'use strict';

const inc = format => {
  format.forEach((element) => {
  if(typeof element === 'number') element ++
  console.log(element);
  })
};

const obj = [1, 2, 'hello',false];

inc(obj);

AlinaDubchak:
// another implementation using map.
'use strict';

const inc = (format) =>{
    let res = format.map(function(element){
    if (typeof element === 'number') element++ 
    return element; 
    })
    return res;
}

const obj = [1, 2, 'hello', false, 4, 'mouse'];

console.log(inc(obj));

PR #40, opened 2022-01-03
Exam

konik56:


Tasks/distinct.js
```diff
@@ -1,16 +1,16 @@
...
-        } else {
-        A.add(a)
-        };
-        w++;
-    });
-  return data;}
+const leaveUniqueValues = function(array) {
+  if (!Array.isArray(array)) {
+    return;
+  };
```
tshemsedinov, line 7:
```suggestion
  if (!Array.isArray(array)) return;
```

Tasks/merge.js
```diff
@@ -1,19 +1,32 @@
...
-        object_3[attribute_name] = object_1[attribute_name]
-  }
-   for (attribute_name in object_2) {
-      (object_3[attribute_name] == object_2[attribute_name])
-     object_3[attribute_name] = object_2[attribute_name]
-    }
-return object_1, object_2, object_3;
-})
+const mergeTwoObjects = function(objFirst, objSecond) {
+  let objFinal = {};
```
tshemsedinov, line 5: Why not const?

Tasks/merge.js
```diff
@@ -1,19 +1,32 @@
...
 
-const result = merge_two_objects({ a: 'uno', b: 'due' }, { c: 'tre' });
+  for (const key in objFirst) {
```
tshemsedinov, line 7: Prefer for..of

Tasks/merge.js
```diff
@@ -1,19 +1,32 @@
...
+    };
+  };
+
+  return objFinal;
+};
+
+const result = mergeTwoObjects({ a: 'uno', b: 'due' }, { c: 'tre' });
 console.log(result);
+
+//Way what is not supported with some browsers
```
tshemsedinov, line 25: Supported everywhere

Tasks/merge.js
konik56: my mistake, forgot about meaning of const is we have link

Tasks/merge.js
konik56: (for (const key of Object.keys(objFirst)) {)- like this ?

PR #41, opened 2022-01-03
Exams

vita133:
made changes in plane.js and difference.js

### DIFF
--- a/Tasks/difference.js
+++ b/Tasks/difference.js
@@ -4,12 +4,12 @@
 const difference = (object_1, object_2) => {
     for (const attribute_name in object_1) {
       object_1[attribute_name] = object_1[attribute_name];
-    }{}
+    }
     if (object_1 in object_2) return false;
     for (const attribute_name in object_2) {
       object_1[attribute_name] = object_2[attribute_name];
       delete object_1[attribute_name];
-    }{}
+    }
     return object_1;
   };
 const result = difference({ a: "uno", b: "due" }, { a: "uno", c: "tre" });
--- a/Tasks/difference.js
+++ b/Tasks/difference.js
@@ -2,12 +2,12 @@
 
 'use strict';
 const difference = (obj1, obj2) => {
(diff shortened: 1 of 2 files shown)


Tasks/difference.js
```diff
@@ -1,16 +1,13 @@
...
-    object_1[attribute_name] = object_2[attribute_name]
-    delete object_1[attribute_name]
-    }{}
-    return object_1
-}
-
-const result = diff({ a: 'uno', b: 'due' }, { a: 'uno', c: 'tre' });
+"use strict";
+const difference = (obj1, obj2) => {
+    let diffObj = { ...obj1 };
```
tshemsedinov, line 5: Prefer const

Tasks/difference.js
```diff
@@ -1,16 +1,13 @@
...
+    for (const key in obj2) {
```
tshemsedinov, line 6: Prefer for..of

Tasks/difference.js
```diff
@@ -1,16 +1,13 @@
...
+      diffObj[key] = obj2[key];
+      delete diffObj[key];
+    }
+    return diffObj;
```
tshemsedinov, line 10: Don't change incoming parameters

Tasks/plane.js
```diff
@@ -1,19 +1,17 @@
...
-    arr[i] = res[j - 1]
-    res.push(value);
-    };
-  };
-  return res
-}
+"use strict";
+const plane = (arr) => {
+  const res = [];
+  for (const value of arr) {
```
tshemsedinov, line 6: for..of over and object?

Tasks/difference.js
tshemsedinov: - Use eslint and prettier
- Respect 2 space indentation

Tasks/plane.js
vita133: >  what do you mean by "object"?

PR #42, opened 2022-01-03
Exams

NikitaSutulov:
Files 'except.js' and 'intersection.js' refactored.

### DIFF
--- a/Tasks/except.js
+++ b/Tasks/except.js
@@ -1,13 +1,17 @@
-// Copy all values from dict except listed
 'use strict';
+
+// Function except: 
+// Returns a new dictionary with all the keys and values from the incomingValues dictionary
+// except for the keys which are in the exceptions rest arguments array.
 const except = (incomingValues, ...exceptions) => {
   const KEYS = Object.keys(incomingValues);
+  const outcomingValues = {};
   KEYS.forEach((key) => {
-    if (exceptions.includes(key)) { 
-      delete incomingValues[key];
+    if (!exceptions.includes(key)) { 
+      outcomingValues[key] = incomingValues[key];
     }
   });
-  return incomingValues;
+  return outcomingValues;
 };
 
 const result = except({ a: 1, b: 'two', c: 3, d: 4 }, 'a', 'd');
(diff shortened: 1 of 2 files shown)


Tasks/except.js
```diff
@@ -1,13 +1,23 @@
...
-   [].sort(() => 2000);
-if (no.includes(Z)) { delete incomingvaluesarray[Z]; return;} else { return; delete incomingvaluesarray[Z]; }
-  }, 2000);
-  ({ key: 'value' });
-  return incomingvaluesarray}
+// Function except: 
+// Returns a new dictionary with all the keys and values from the incomingValues dictionary
+// except for the keys which are in the exceptions rest arguments array.
+const except = (incomingValues, ...exceptions) => {
+  const KEYS = Object.keys(incomingValues);
```
tshemsedinov, line 7: `camelCase`

Tasks/except.js
```diff
@@ -1,13 +1,23 @@
...
+  const outcomingValues = {};
+  KEYS.forEach((key) => {
+    if (!exceptions.includes(key)) { 
+      outcomingValues[key] = incomingValues[key];
+    }
+  });
+  return outcomingValues;
```
tshemsedinov, line 14: Don't change incoming parameters

Tasks/intersection.js
```diff
@@ -1,17 +1,22 @@
...
-delete object_1[attribute_name];                       {
-                                                       }
-                  }{}
-                  }{}
-return object_1;
-}
+// Function intersection
+// Returns a new object with all the keys and values which are in both object1 and object2
+const intersection = (object1, object2) => {
+  const OBJECT1_KEYS = Object.keys(object1);
```
tshemsedinov, line 6: `camelCase`

Tasks/except.js
NikitaSutulov: The function (except) does not modify incoming parameters, it creates a new dictionary (outcomingValues), in which there are the keys from the incoming dictionary (incomingValues) which are not in the rest parameters array (...exceptions), and the values of these keys. And the incomingValues and ...exceptions parameters are not modified at all.

tshemsedinov:
forEach have very limited usage in clean code, you can safely use it just without outer structures mutations, for examle: `connection.forEach((connection) => connection.close());`

PR #43, opened 2022-01-03
Exam

neliudochka:
Files 'difference.js' and 'distinct.js' refactored

### DIFF
--- a/Tasks/difference.js
+++ b/Tasks/difference.js
@@ -1,7 +1,7 @@
 // Find a difference of two dictionaries
 'use strict';
 
-const dictDifference = ({...obj1}, obj2) => {
+const dictDifference = ({ ...obj1 }, obj2) => {
   for (const key in obj2) {
     delete obj1[key];
   }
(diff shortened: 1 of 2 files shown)


Tasks/difference.js
```diff
@@ -1,16 +1,18 @@
...
-    object_1[attribute_name] = object_2[attribute_name]
-    delete object_1[attribute_name]
-    }{}
-    return object_1
-}
+const dictDifference = ({ ...obj1 }, obj2) => {
+  for (const key in obj2) {
+    delete obj1[key];
+  }
+  return obj1;
```
tshemsedinov, line 8: Don't change incoming parameters

Tasks/difference.js
```diff
@@ -1,16 +1,18 @@
...
-    object_1[attribute_name] = object_1[attribute_name]
-    }{}
-    if (object_1 in object_2) return false
-    for (attribute_name in object_2) {
-    object_1[attribute_name] = object_2[attribute_name]
-    delete object_1[attribute_name]
-    }{}
-    return object_1
-}
+const dictDifference = ({ ...obj1 }, obj2) => {
```
tshemsedinov, line 4: Why clone just first object?

tshemsedinov: And it isn't clear/obvious syntax for common use

Tasks/difference.js
```diff
@@ -1,16 +1,18 @@
...
+  for (const key in obj2) {
```
tshemsedinov, line 5: Prefer for..of

PR #46, opened 2022-01-03
Exams

MrSampy:


Tasks/Replace.cpp
```diff
@@ -0,0 +1,17 @@
...
+  {
+    return str;
+  }
+  return str.erase(positionofsubstr, substr.size()).insert(positionofsubstr, newstr);
+}
+int main()
+{
+  cout << replace("Hello <username> and bye!", "<username>", "Marcus");
+  return 0;
+}
```
tshemsedinov, line 17: Add \n at EOF

Tasks/Replace.cpp
```diff
@@ -0,0 +1,17 @@
+#include <iostream>
+#include <string>
+using namespace std;
+const string replace(string str = "", const string substr = "", const string newstr = "")
+{
+  const int positionofsubstr = str.find(substr);
+  if (substr == "" || positionofsubstr == -1)
+  {
+    return str;
+  }
```
tshemsedinov, line 10:
```suggestion
  if (substr == "" || positionofsubstr == -1) return str;
```

tshemsedinov: It's better to check `substr == ""` even before `str.find(substr);`

Tasks/words.cpp
```diff
@@ -0,0 +1,33 @@
...
+      flag = false;
+    }
+    else
+    {
+      if (!flag)
+      {
+        numberofwordsinstr++;
+      }
+      flag = true;
+    }
```
tshemsedinov, line 24: Trash... plz rewrite more compact, is it a Pascal habit?

PR #47, opened 2022-01-03
Exams

yarman1:


yarman1 (commented):
Made a terminological mistake in some commits. There should be 'function parameters' instead of 'arguments'.

Tasks/except.js
```diff
@@ -1,13 +1,14 @@
...
-EXCEPT =(incomingvaluesarray,...no)=> {
- intermediate_variable = Object.keys(incomingvaluesarray, 'a', 'b', 'c');
-   intermediate_variable.forEach((Z) => {
-   [].sort(() => 2000);
-if (no.includes(Z)) { delete incomingvaluesarray[Z]; return;} else { return; delete incomingvaluesarray[Z]; }
-  }, 2000);
-  ({ key: 'value' });
-  return incomingvaluesarray}
+const except = (dict, ...exceptions) => {
+  const cleanDict = {};
```
tshemsedinov, line 5: Probably we can improve naming. `clean` rly? I'd prefer `dict` or even `result`

tshemsedinov (commented):
:+1:

PR #48, opened 2022-01-03
Exams

GddgdgMen:


Tasks/difference.js
```diff
@@ -1,16 +1,11 @@
...
-    object_1[attribute_name] = object_1[attribute_name]
-    }{}
-    if (object_1 in object_2) return false
-    for (attribute_name in object_2) {
-    object_1[attribute_name] = object_2[attribute_name]
-    delete object_1[attribute_name]
-    }{}
-    return object_1
+const diff = (object1, object2) => {
+  for (const key in object2) {
```
tshemsedinov, line 5: Prefer for..of here

Tasks/difference.js
```diff
@@ -1,16 +1,11 @@
...
+    delete object1[key];
+  }
+  return object1;
```
tshemsedinov, line 8: Don't change incoming parameters

Tasks/intersection.js
```diff
@@ -1,17 +1,16 @@
...
+    if (name !== "" && object2[name]) {
+        object2[name] = object1[name];
+    } else {
+      delete object1[name];
+    }
+  }
+  return object1;
+};
+const result = intersection({ a: "uno", b: "due" }, { a: "uno", c: "tre" });
+console.log(result);
```
tshemsedinov, line 16: Add \n at EOF

tshemsedinov (commented):
- Use eslint and prettier

GddgdgMen:
Updated the code

PR #49, opened 2022-01-03
Exams

DaniilAlyeksyeyenko:


Tasks/intersection.js
```diff
@@ -1,16 +1,17 @@
...
-} for (attribute_name of first_keys)                   {
-if (attribute_name !== '' && object_2[attribute_name]) {
-object_2[attribute_name] = object_1[attribute_name];   {
-}} else                                                {
-delete object_1[attribute_name];                       {
-                                                       }
-                  }{}
-                  }{}
-return object_1;
+function intersection(first, second) {
```
tshemsedinov, line 4: Prefer arrow functions

Tasks/intersection.js
```diff
@@ -1,16 +1,17 @@
...
+  const firstKeys = Object.keys(first);
+  const res = {};
+  for (const attributeName of firstKeys) {
+    if (attributeName !== '' && second[attributeName]) {
+      res[attributeName] = first[attributeName];
+    } else {
+      delete res[attributeName];
+    }
+  }
+  return res;
```
tshemsedinov, line 14: Don't change incoming parameters

Tasks/words.js
```diff
@@ -1,26 +1,16 @@
...
-   }
- }
- return number_of_words_in_s
- }
+function words(str) {
+  let countWords = 0;
+  let flag = false;
+  for (const char of str) {
+    if (!flag && char !== ' ') { countWords++; }
+    char === ' ' ? flag = false : flag = true;
```
tshemsedinov, line 10: - Use eslint and prettier
- Use if-statement instead ternary in cases like this
- But in this certain case you can assign boolean result of comparison directly to the flag

PR #50, opened 2022-01-03
Exams

DAN1ROCK:
Refactored 'merge.js' and 'difference.js'


tshemsedinov (commented):
- Use eslint and prettier
- Prefer for..of in this case

tshemsedinov:
Don't change incoming parameters

PR #55, opened 2022-01-03
Exams

Ivanopolus21:


Tasks/skip.js
```diff
@@ -1,17 +1,21 @@
...
+
 // Remove values from array
 
-Skip =(T,...Remove)=> {
-x = 0;
-for (C of T) {
-for (X of Remove) {
-if (C === X) {
-  T.splice(x, 1);
+const skip = (array, ...charsToRemove) => {
```
tshemsedinov, line 5: Why chars?

Tasks/words.js
```diff
@@ -1,26 +1,22 @@
...
-   if (!flag) {
-   if (c === ' ') {
-   if (flag === true) { flag = false } else { flag = false }
-    } else {
-   if (flag === true) { flag = true } else { flag = true }
-      number_of_words_in_s++
+const words = (string) => {
+  let numberOfWords = 0;
+  let flag = false;
+  for (const element of string) {
```
tshemsedinov, line 8: Here `char` will be ok )

Tasks/skip.js
```diff
@@ -1,17 +1,21 @@
...
+  let removedChar = 0;
+  for (const selectedChars of array) {
+    for (const selectedCharsToRemove of charsToRemove) {
```
tshemsedinov, line 8: Try to avoid nested loops

Tasks/skip.js
Ivanopolus21: Was changed to values!

Tasks/words.js
Ivanopolus21: Good idea, thank You! Was changed!

Tasks/skip.js
Ivanopolus21: Got it! Tried to remove one loop and use some methods instead

PR #56, opened 2022-01-03
Exams

junkjuk:


Tasks/plane.js
```diff
@@ -1,19 +1,13 @@
...
-    res.push(...const_plane(value));
-    arr[i] = res[i - 1]
-    }else{
-    arr[i] = res[j - 1]
-    res.push(value);
-    };
-  };
-  return res
-}
+const plane = arr => arr.flatMap( element => {
```
tshemsedinov, line 3: Use round brackets even for single argumanet lambda functions

Tasks/plane.js
```diff
@@ -1,19 +1,13 @@
...
+    if (Array.isArray(element)) {
+      return plane(element);
+    } else {
+      return element;
+    }
```
tshemsedinov, line 8: Do not need else here

Tasks/words.js
```diff
@@ -1,26 +1,18 @@
...
-   if (!flag) {
-   if (c === ' ') {
-   if (flag === true) { flag = false } else { flag = false }
-    } else {
-   if (flag === true) { flag = true } else { flag = true }
-      number_of_words_in_s++
+const words = str => {
+  let wordsNumber = 0;
+  let isWord = false;
+  for (let element of str) {
```
tshemsedinov, line 6: Prefer const

Tasks/words.js
```diff
@@ -1,26 +1,18 @@
...
-      if (flag === true) { flag = false } else { flag = false }
-    } else {
-      if (flag === true) { flag = true } else { flag = true }
-    }
-   }
- }
- return number_of_words_in_s
- }
+  }
+return wordsNumber;
```
tshemsedinov, line 14: Respect 2 space indentation

tshemsedinov (commented):
Use eslint and prettier

Tasks/words.js
junkjuk: I also prefer const, but I can't insert it here because then I do wordsNumber++ and isWord = true, and with const it can't be done. I can create an object with fields wordsNumber and isWord, and then refer to them through the object, but I don't think it's a good practice

junkjuk: Replaced let element to const in this commit 47f5c58

PR #60, opened 2022-01-03
Exam

DaniloYaremenko:


tshemsedinov (commented):
- Use eslint and prettier
- Improve code style
- Respect 2 space indentation
- Remove unneeded operations, calls, arguments, blocks, etc.
- Optimize lexical scope, minimize area of identifier visibility

PR #62, opened 2022-01-03
Exams

michigang1:
I translated on Kotlin and refactored 2 works: сompare.js and replace.js


Kotlin/src/replace.kt
```diff
@@ -0,0 +1,24 @@
+fun replace( string: String, subString: String, newString: String): String{
+    if(subString == "") return string
+    else{
```
tshemsedinov, line 3: We don't need `else` here

Kotlin/src/replace.kt
```diff
@@ -0,0 +1,24 @@
...
+        var stringElse = string
+        var resultElse = ""
+        do{
+            val index = stringElse.indexOf(subString)
+            if(index == -1){
+                resultElse += stringElse
+                return resultElse
+            }
+            else{
```
tshemsedinov, line 12: Ditto

Kotlin/src/replace.kt
```diff
@@ -0,0 +1,24 @@
+fun replace( string: String, subString: String, newString: String): String{
+    if(subString == "") return string
+    else{
+        var stringElse = string
+        var resultElse = ""
+        do{
```
tshemsedinov, line 6: I'd prefer `while`

Kotlin/src/replace.kt
```diff
@@ -0,0 +1,24 @@
...
+                stringElse = stringElse.substring(indexSubString,stringElse.length)
+                resultElse += subStringElse + newString
+            }
+        }while (true)
+    }
+}
+fun main(){
+    val result = replace("Hello <username> and bye!", "<username>", "Marcus")
+    println(result)
+}
```
tshemsedinov, line 24: Add \n at EOF

Kotlin/src/replace.kt
```diff
@@ -0,0 +1,24 @@
...
+                resultElse += stringElse
+                return resultElse
+            }
+            else{
+                val subStringElse = stringElse.substring(0,index)
+                val indexSubString = index + subString.length
+                stringElse = stringElse.substring(indexSubString,stringElse.length)
+                resultElse += subStringElse + newString
+            }
+        }while (true)
```
tshemsedinov, line 18: Strange code alignment

Kotlin/src/compare.kt
```diff
@@ -0,0 +1,16 @@
...
+fun <K,V> compare(firstMap: Map<K,V>, secondMap: Map<K,V>): Boolean {
+    val keysOfFirstMap = firstMap.keys
+    val keysOfSecondMap = secondMap.keys
+    var compareBool = true
+    return if (keysOfFirstMap.size == keysOfSecondMap.size) {
+        for (key in keysOfFirstMap)
+            if (firstMap[key] != secondMap[key]) compareBool = false
+         compareBool
+    }
+    else compareBool
```
tshemsedinov, line 11: Use `return early` principle, it would be much better readable than `return if`

PR #66, opened 2022-01-03
Exams 

jinworldwildhandsome:


tshemsedinov (commented):
- Use eslint and prettier
- Respect 2 space indentation
- It's better to rename `key1` and `key2` to `keys1` and `keys2`

jinworldwildhandsome:
- used eslint and prettier for both tasks; 
- renamed key1 and key2 to keys1 and keys2 for merge.js

PR #72, opened 2022-01-04
Exams

SavageVictor:


Tasks/password.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        private static Random _random = new Random();
+        
+        public static void Main()
+        {
+            string result = GeneratePassword("abc123", 7);
+            Console.WriteLine(result);
+        }
+
+        private static string GeneratePassword(string alphabet, int passwordLength)
```
tshemsedinov, line 15: I'd prefer lambda function here

Tasks/password.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            int maxAlphabetIndex = alphabet.Length;
+            string password = "";
+            for (int i = 0; i < passwordLength; i++)
+            {
+                int index = (int) Math.Floor(_random.NextDouble() * maxAlphabetIndex);
```
tshemsedinov, line 21: Decompose: extract `int random(in min, int max)`

Tasks/password.cs
```diff
@@ -0,0 +1,28 @@
...
+                password += alphabet[index];
+            }
+
+            return password;
+        }
+    }
+}
```
tshemsedinov, line 28: Add \n at EOF

Tasks/size.cs
```diff
@@ -0,0 +1,48 @@
...
+                    return fileSize / 1000 + " kb";
+                }
+                else if(exp == 2)
+                {
+                    return fileSize / 1000000 + " mb";
+                }
+                else if(exp >= 3)
+                {
+                    return fileSize / 1000000000 + " gb";
+                }
```
tshemsedinov, line 42: Use collection for converting

PR #73, opened 2022-01-04
Exams

jugular237:
Тимур Гафарович, смотрите коммиты начиная с FIRST COMMIT.


tasks1(difference)/Program1.cs
```diff
@@ -0,0 +1,14 @@
+﻿    Dictionary<Tvalue1, Tvalue2> Difference<Tvalue1, Tvalue2>
+        (Dictionary<Tvalue1, Tvalue2> dict1, Dictionary<Tvalue1, Tvalue2> dict2)
+    {
+        foreach (Tvalue1 dictKey in dict2.Keys)
+            dict1.Remove(dictKey);
```
tshemsedinov, line 5: Don't change incoming parameters

tshemsedinov (commented):
Don't commit a bunch of crap like json and xml files with code, it's not a project, it's just an exams so searching for 3 lines of code in an ocean of boilerplate is not I am expected for this new year )))

tasks1(difference)/Program1.cs
jugular237: Are you talking about type of parameters or parameters names?If about type then i dont know on which type i must replace dictionary type, cause dictionary replaces js objects the best in this case.

PR #74, opened 2022-01-04
Exams

HoloborodkoM:


Tasks/between.js
```diff
@@ -1,18 +1,17 @@
...
-      str =  str.substring(k)
-      if ( s) {
-        i = str.indexOf( s )
-        if (i === -1) {
-          return  '' } else { str =  str.substring(0 , i)
-            }}}
-  return str
-}
+const getValueBetween = (string, prefix, suffix) => {
+  const ERR = -1;
```
tshemsedinov, line 5: Use `camelCase`

Tasks/between.js
```diff
@@ -1,18 +1,17 @@
...
+  let substring = '';
+  const prefixIndex = string.indexOf(prefix);
+  const suffixIndex = string.indexOf(suffix);
+  if (prefixIndex !== ERR && suffixIndex  !== ERR) {
```
tshemsedinov, line 9: Use eslint and prettier to fix code style and common conventions

Tasks/difference.js
```diff
@@ -1,16 +1,19 @@
...
-    delete object_1[attribute_name]
-    }{}
-    return object_1
-}
+const diff = (objectFirst, objectSecond) => {
+  const objectDiff = { ...objectFirst };
+  for (const attributeName in objectSecond) {
+    const valueObjFirst = objectFirst[attributeName];
+    const valueObjSecond = objectSecond[attributeName];
+    if (typeof(valueObjSecond) !== 'undefined') {
```
tshemsedinov, line 9: Typeof of an operator, not a function

Tasks/difference.js
```diff
@@ -1,16 +1,19 @@
...
-    if (object_1 in object_2) return false
-    for (attribute_name in object_2) {
-    object_1[attribute_name] = object_2[attribute_name]
-    delete object_1[attribute_name]
-    }{}
-    return object_1
-}
+const diff = (objectFirst, objectSecond) => {
+  const objectDiff = { ...objectFirst };
+  for (const attributeName in objectSecond) {
```
tshemsedinov, line 6: Prefer for..of

Tasks/between.js
HoloborodkoM: I used only eslint.

PR #77, opened 2022-01-04
Exams

yaryna-bashchak:


Tasks/merge.js
```diff
@@ -1,19 +1,19 @@
...
-   for (attribute_name in object_1) {
-       (
-          object_3[attribute_name] == object_1[attribute_name]
-       );
-        object_3[attribute_name] = object_1[attribute_name]
+const merge = (object1 = {}, object2 = {}) => {
+  const object3 = { ...object1 };
+  const keys2 = Object.keys(object2);
+  for (const key of keys2) {
+    if (!(key in object3))
```
tshemsedinov, line 9: Better use `includes`

Tasks/plane.js
```diff
@@ -1,19 +1,21 @@
...
-    res.push(...const_plane(value));
-    arr[i] = res[i - 1]
-    }else{
-    arr[i] = res[j - 1]
-    res.push(value);
-    };
-  };
-  return res
-}
+const plane = array => {
```
tshemsedinov, line 5: Use round brackets even for single argumanet lambda functions

Tasks/merge.js
yaryna-bashchak: > Better use `includes`

okay, I fixed it

Tasks/plane.js
yaryna-bashchak: > Use round brackets even for single argumanet lambda functions

okay, I fixed it

PR #80, opened 2022-01-04
Exams

crazysparrow69:


Tasks/compare.js
```diff
@@ -1,19 +1,23 @@
...
-    if (first_values[c] === second_values[c]) e = e && true
-   else { e = e && false }
-   }
-  return e;
-  }
-)
+const compareDictionaries = function ( obj1, obj2 ) {
+    const obj1Keys = Object.keys(obj1).sort();
+    const obj2Keys = Object.keys(obj2).sort();
+    let e = true;
```
tshemsedinov, line 7: Try to improve naming

Tasks/compare.js
```diff
@@ -1,19 +1,23 @@
...
+    
+    if ( obj1Keys.length !== obj2Keys.length ) return false;
```
tshemsedinov, line 9: Use eslint and prettier to check style against common conventions
```suggestion
    if (obj1Keys.length !== obj2Keys.length) return false;
```

Tasks/compare.js
```diff
@@ -1,19 +1,23 @@
...
+    
+    for ( let i = 0; i < obj1Keys.length; i++ ) {
+        if ( obj1Keys[i] !== obj2Keys[i] ) return false; 
+    }
+    
+    for (const key of obj1Keys) {
+        if (obj1[key] !== obj2[key]) e = false;
+    }
+    
+    return e;
```
tshemsedinov, line 19: Do we need a flag at all?

Tasks/compare.js
```diff
@@ -1,19 +1,23 @@
...
+    
+    if ( obj1Keys.length !== obj2Keys.length ) return false;
+    
+    for ( let i = 0; i < obj1Keys.length; i++ ) {
+        if ( obj1Keys[i] !== obj2Keys[i] ) return false; 
+    }
+    
+    for (const key of obj1Keys) {
+        if (obj1[key] !== obj2[key]) e = false;
+    }
```
tshemsedinov, line 17: We can easily combine two loops

PR #82, opened 2022-01-04
Exams

GAlexandrD:


GAlexandrD:
alternative implementation of distinct:
`const distinct = (array) => Array.from(new Set(array));`
it seems to me that plane can be optimized using closure:
```
const plane = (array) => {
  const result = [];
  const closedPlane = (array) => {
    for (const item of array) {
      if (Array.isArray(item)) closedPlane(item);
      else result.push(item);
    }
    return result;
  };
  return closedPlane(array);
};
```
We don't create temporary arrays and don't use unnecessary rest operators and push calls. But option with closure is larger.

tshemsedinov:
- Prefer for..of

tshemsedinov:
What is easy to read recursion with closure or just a recursion?

GAlexandrD:
> * Prefer for..of
> * Don't change incoming parameters

i don't understand where i change parameter? I copied it in the result.

Tasks/distinct.js
```diff
@@ -1,16 +1,14 @@
...
-        if (A.has(a)) {
-        data.splice(w, 1)
-        } else {
-        A.add(a)
-        };
-        w++;
-    });
-  return data;}
+const distinct = (array) => {
+  const result = [...array];
```
GAlexandrD, line 4: > * Prefer for..of
> * Don't change incoming parameters

I changed result, is it parameter? this way create new array and copy each element, I tried array === result and it showed false.

PR #83, opened 2022-01-04
Exam completed

svitbar:


Tasks/duplicate.js
```diff
@@ -1,15 +1,16 @@
...
 
-duplicate = (value, N) => {
-  if (N <= 0) return [];
-  else {
-    res = [];
-    for (let i = 0; i < N; i++) {
-      res[i] = value;
+const duplicateValue = (value, times) => {
+  const res = [];
+  if (times >= 0) {
```
tshemsedinov, line 7: Use return-early principle

Tasks/duplicate.js
```diff
@@ -1,15 +1,16 @@
...
+
 // Push single value multiple times
 
-duplicate = (value, N) => {
-  if (N <= 0) return [];
-  else {
-    res = [];
-    for (let i = 0; i < N; i++) {
-      res[i] = value;
+const duplicateValue = (value, times) => {
```
tshemsedinov, line 5: I'd prefer to add default value for `times`

PR #84, opened 2022-01-04
Exams

dianashamota:


Tasks/difference.js
```diff
@@ -1,16 +1,15 @@
...
-    }{}
-    if (object_1 in object_2) return false
-    for (attribute_name in object_2) {
-    object_1[attribute_name] = object_2[attribute_name]
-    delete object_1[attribute_name]
-    }{}
-    return object_1
-}
+const difference = (object1, object2) => {
+  const diffDict = Object.assign({}, object1);
```
tshemsedinov, line 6: That's of nut plz rewrite without Object.assign

PR #86, opened 2022-01-04
Exam

EvhenyiAlekseevych:


intersection.js
```diff
@@ -0,0 +1,19 @@
+// Finde an intersection of two dictionaries
+'use strict';
+const dictIntersection = (firstDict, secDict) => {
+  const firstDictKeys = Object.keys(firstDict);
+  const result = {};
+  for (const element of firstDictKeys) {
+    if (firstDict[element] === secDict[element]) {
+      result[element] = firstDict[element];
```
tshemsedinov, line 8: I'd prefer to use intermediate variables

sum.js
```diff
@@ -0,0 +1,15 @@
+// Sum all numbers from array
+'use strict';
+const sumNumbers = (array) => {
```
tshemsedinov, line 3: Add an empty line before to separate visually

PR #88, opened 2022-01-04
Exams

nikitaq121:


tshemsedinov:
Use eslint and prettier, follow all style conventions, fix all warnings
Try to optimize more

PR #90, opened 2022-01-04
Exams

Mikkador:


Tasks/shuffle.js
```diff
@@ -1,9 +1,10 @@
 // Shuffle an array
+'use strict';
+const shuffle = (array) => {
+  const CHANCE_EQUALISER = 0.5;
```
tshemsedinov, line 4: Move out global constant

Tasks/shuffle.js
```diff
@@ -1,9 +1,10 @@
...
+  array.sort(() => Math.random() - CHANCE_EQUALISER); 
+  return array;
```
tshemsedinov, line 6: Don't change incoming parameters

PR #92, opened 2022-01-04
Exams

romalitv:


tshemsedinov:
Don not use `console.log` in functions, that is just for logging/debug purposes. If you need to return an error you can use `return` operator with `new Error` or error code

tshemsedinov:
Use eslint and prettier

PR #94, opened 2022-01-04
Exams

m1guelten:
Potapenko Misha IM-13


tshemsedinov:
- Use eslint and prettier
- Prefer const
- Try to use collection instead of multiple if-statements

PR #95, opened 2022-01-04
Exams

thesemik:
Semivolos Eugene IM-11

### DIFF
--- a/Tasks/compare.js
+++ b/Tasks/compare.js
@@ -6,12 +6,10 @@ const compare = (list1, list2) => {
   const key1 = Object.keys(list1);
   const key2 = Object.keys(list2);
   if (key1.join() !== key2.join()) return false;
-  let e = true;
   for (let key of key1) {
-    if (list1[key] === list2[key]) e = e && true;
-    else { e = e && false; }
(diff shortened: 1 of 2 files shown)


Tasks/compare.js
```diff
@@ -1,19 +1,16 @@
...
-   let e = true;
-   for (c of a) {
-    if (first_values[c] === second_values[c]) e = e && true
-   else { e = e && false }
-   }
-  return e;
+const compare = (list1, list2) => {
+  const key1 = Object.keys(list1);
+  const key2 = Object.keys(list2);
+  if (key1.join() !== key2.join()) return false;
```
tshemsedinov, line 8: Think about .sort()

Tasks/plane.js
```diff
@@ -1,19 +1,17 @@
...
-  };
-  return res
-}
+const plane = (arr) => {
+  let res = [];
+  for (let value of arr) {
+    const par = Array.isArray(value)
+      ? plane(value)
+      : [value]; 
+    res.push(...par);
```
tshemsedinov, line 11: if statement will be bettr here

PR #96, opened 2022-01-04
Exams

SergiiTkachuk:


tshemsedinov:
- Use eslint and prettier
- Prefer for..of

PR #100, opened 2022-01-05
Exams

YanPetrov7:
Ян Петров IM-12

### DIFF
--- a/Tasks/between.js
+++ b/Tasks/between.js
@@ -2,21 +2,21 @@
 'use strict';
 
 const getValueBetween = (str, fromLeft, fromRight) => {
-  let index = str.indexOf(fromLeft)
-    if ( index === -1 ){
-       return ''
+  let index = str.indexOf(fromLeft);
+  if (index === -1) {
+    return '';
+  } else {
+    index += fromLeft.length;
+    str = str.substring(index);
+    index = str.indexOf(fromRight);
+    if (index === -1) {
+      return '';
     } else {
-      index += fromLeft.length
-      str =  str.substring(index)
-        index = str.indexOf(fromRight)
-        if (index === -1) {
-          return  '' 
-        } else { 
-          str =  str.substring(0 , index)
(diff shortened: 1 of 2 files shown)


Tasks/between.js
```diff
@@ -1,18 +1,22 @@
...
-        if (i === -1) {
-          return  '' } else { str =  str.substring(0 , i)
-            }}}
-  return str
-}
+const getValueBetween = (str, fromLeft, fromRight) => {
+  let index = str.indexOf(fromLeft);
+  if (index === -1) {
+    return '';
+  } else {
```
tshemsedinov, line 8: Use return-early principle and we don't need else here

Tasks/between.js
```diff
@@ -1,18 +1,22 @@
...
+    index += fromLeft.length;
+    str = str.substring(index);
+    index = str.indexOf(fromRight);
+    if (index === -1) {
+      return '';
+    } else {
+      str = str.substring(0, index);
+    }
```
tshemsedinov, line 16: Ditto

Tasks/words.js
```diff
@@ -1,26 +1,27 @@
...
+const countWords = (str) => {
+  let wordsNum = 0;
+  let flag = false;
+  for (const element of str) {
+    if (!flag) {
+      if (element === ' ') {
+        flag = false;
+      } else {
+        flag = true;
+      }
```
tshemsedinov, line 13: Try to optimize this

Tasks/between.js
YanPetrov7: I thinked it will change the code style, but okay.

Tasks/between.js
YanPetrov7: Ditto

Tasks/words.js
YanPetrov7: Okay, you can see the changings in a new commit.

PR #101, opened 2022-01-05
Exams

mikhmol:


tshemsedinov:
- Dont move `days` into function, it is a constant
- Try to rewrite with shorter expressions optimized for readability
- Windows, rly? Why don't use DOS?

mikhmol:
> * Dont move `days` into function, it is a constant
Moved out from function
> * Try to rewrite with shorter expressions optimized for readability
Rewrited. Shall be more optimal because case conversition for input argument moved out from a loop.
> * Windows, rly? Why don't use DOS?
Sorry, will use linux in future

mikhmol:
I switched to macOS

PR #102, opened 2022-01-05
Exams

Sviat1001:


tshemsedinov:
- Use eslint and prettier
- Respect 2 space indentation
- Use for..of instead for..in
- Prefer arrow functions
- Never use `==`
- Semicolons?

PR #103, opened 2022-01-05
Exams

tomatorepublic:


tshemsedinov:
- Improve naming
- Use eslint and prettier
- Try to use `includes` method instead of `indexOf`
- If we use block in one branch of if-statement we should use it for another branch too
- Prefer arrow functions
- Don't change `days` from function

tomatorepublic:
> > * Try to use `includes` method instead of `indexOf`

But if this is not a high-load application, then maybe we could use regexp.
Can we use it? This way the code will be smaller and more understandable.

tshemsedinov:
`str.substring(str.indexOf(prefix)+prefix.length, str.indexOf(suffix))` is ok but it's better to use intermediate variables:
```js
const start = str.indexOf(prefix) + prefix.length;
const end = str.indexOf(suffix);
const result = str.substring(start, end);
```

tshemsedinov:
Avoid regexp, it can bring security and performance problems

PR #104, opened 2022-01-05
Exams

vadumpastushok:


tshemsedinov:
- Try to rewrite with loops, split/join is a hack, not a good way
- Use eslint and prettier
- Respect `camelCase`

tshemsedinov:
- Use eslint and prettier
- Do we need `EMPTY`?
- Solution with split/map is a hack, try to rewrite
- split/join is also not optimal way

PR #105, opened 2022-01-05
Exam

Snare295:
Neumitiy Vlaislav IM-13


tshemsedinov:
- Use eslint and prettier
- Try rewrite without switch/case

PR #106, opened 2022-01-05
Exams

lekoaa:


tshemsedinov:
- Common Python naming convention is `snake_case`
- Add empty lines for visual separation
- Add \n at EOF
- Do not use pass-through parameters

lekoaa:
Done ^_^

PR #121, opened 2022-01-15
Exams

nieopierzony:
- distinct
- drop
- difference:
  In the original code, the answer does not show the value of c, which was added in the second object. I thought it was a logical error, because it also refers to the difference
- duplicate:
    I really did not understand at first what this function was supposed to do, reading only its name and the comment, which is copied from another task. It seemed to me that the name fillArray fits better here

### DIFF
--- a/Tasks/difference.js
+++ b/Tasks/difference.js
@@ -1,17 +1,17 @@
 'use strict';
 
 // Find a difference between two dictionaries
-const diff = (object_1, object_2) => {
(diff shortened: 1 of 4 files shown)


tshemsedinov:
I'll provide tests to check functionality and special analyzer to generate hints how to optimize and improve style.
Meanwhile you can check yourself with following hints: https://github.com/HowProgrammingWorks/Exams/blob/master/HINTS.md

PR #124, opened 2022-01-17
Set eslint rule "always" for "arrow-parens"

nieopierzony:
Reference to hints:
https://github.com/HowProgrammingWorks/Exams/blob/72048e5eef6d8bca2c31131ae507ee899d2b7505/HINTS.md?plain=1#L13


tshemsedinov (changes requested):
Do not mix file formatting with rule changes in s single PR

.eslintrc.json
```diff
@@ -265,4 +265,4 @@
       "never"
     ]
   }
-}
+}
```
tshemsedinov, line 268:
```suggestion
}

```

PR #151, opened 2023-01-06
Pull request (Maxim Diachok) for exams

BbIXOD:
1. count.js
2. replace.js

### DIFF
--- a/Tasks/count.js
+++ b/Tasks/count.js
@@ -1,16 +1,13 @@
+'use strict';
 // Sum all number values in dict
 
-count = (obj) => {
-  sum = 0;
-  {
-  }
-  keys = Object.keys(obj);
-  keys.forEach((key) => {
-    [];
-    const value = obj[key];
+const count = obj => {
+  let sum = 0;
+  const values = Object.values(obj);
+
+  for (const value of values)
     if (typeof value === 'number') sum += value;
-  });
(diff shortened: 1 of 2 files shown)


kol-oss (approved):
LGTM. Nothing to add.

Tasks/count.js
```diff
@@ -1,16 +1,13 @@
...
-  sum = 0;
-  {
-  }
-  keys = Object.keys(obj);
-  keys.forEach((key) => {
-    [];
-    const value = obj[key];
+const count = obj => {
+  let sum = 0;
+  const values = Object.values(obj);
```
tshemsedinov, line 6: Alternatively you can use `reduce` to sum values

Tasks/replace.js
```diff
@@ -1,22 +1,25 @@
...
-        src = src.substring(_index + substr.length, src.length);
-        res += start + newstr;
-      }
-    } while (true);
   }
+
+  let source = str;
+  let res = '';
+
+  while (true) {
```
tshemsedinov, line 12: It's even `replaceAll` :smiley:

BbIXOD: I know but we don't have to change realization. Also function, that contains only one line it is weird. I thought that here another realization of replaceAll

PR #153, opened 2023-01-07
Exams by Daniil Timofeev

SharpDevOps10:
I made the general improvement  of the following solutions (Self-esteem - 95)

### DIFF
--- a/Tasks/intersection.js
+++ b/Tasks/intersection.js
@@ -1,16 +1,16 @@
 // Find an intersection of two dictionaries
 'use strict';
-const intersection = (object_1, object_2) => {
-  first_keys = Object.keys(object_1);
+const intersection = (object1, object2) => {
+  const object1Keys = Object.keys(object1);
   {
   }
-  for (attribute_name of first_keys) {
-    if (object_1[attribute_name] === object_2[attribute_name]) {
-      object_2[attribute_name] = object_1[attribute_name];
+  for (const key of object1Keys) {
+    if (object2[key] === object1[key]) {
+      object2[key] = object1[key];
       {
       }
     } else {
-      delete object_1[attribute_name];
+      delete object1[key];
       {
       }
     }
@@ -19,7 +19,7 @@ const intersection = (object_1, object_2) => {
   }
   {
   }
-  return object_1;
+  return object1;
 };
 
 require('../Tests/intersection.js')(intersection);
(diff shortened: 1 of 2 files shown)


basarabst (approved):
No problems were found. Approved.

Tasks/intersection.js
```diff
@@ -1,25 +1,14 @@
...
-      delete object_1[attribute_name];
-      {
-      }
+'use strict';
+const intersection = (object1, object2) => {
+  const object1Keys = Object.keys(object1);
+  const result = {};
+  for (const key of object1Keys) {
+    if (object2[key] === object1[key]) {
+      result[key] = object1[key];
```
tshemsedinov, line 8: You can save `object[key]` to intermediate variable not to repeat execution of the same expression

Tasks/plane.js
```diff
@@ -1,18 +1,16 @@
...
-require('../Tests/plane.js')(const_plane);
+'use strict';
+const planedArray = (arr) => {
+  const res = [];
+  for (const value of arr) {
+    if (Array.isArray(value)) {
+      res.push(...planedArray(value));
+    } else {
+      res.push(value);
+    }
```
tshemsedinov, line 10: Alternatively you can use ternary here, but it's ok

Tasks/plane.js
```diff
@@ -1,18 +1,16 @@
...
+  }
+  return res;
+};
+const result = planedArray([[1, [[2]], [3, 4], [5], [6, [7, 8]]]]);
+console.log(result);
```
tshemsedinov, line 15: Hint: Do not need tests/example here, we call tests from different place

tshemsedinov (approved):

Tasks/intersection.js
SharpDevOps10: Thank you for finding my code`s disadvantages. I`ve got it, thus I`ve made new remastered commits with the following fixes. There is the link : https://github.com/HowProgrammingWorks/Exams/compare/master...SharpDevOps10:Exams:exams
P.S new commits are marked as "remastered " and made on Jan 25, 2023

Tasks/plane.js
SharpDevOps10: P.P.S I`ve changed 'else-if' block into the ternary operator as well. This advice was alternative, hence I`ve marked it as 'optional'

PR #156, opened 2023-01-08
Exams by Mariia Kryvokhata (IM-21)

mari4kaa:
Refactored compare.js and quotes.js ;
Code review was made by @sonakrrr and @a42b

### DIFF
--- a/Tasks/quotes.js
+++ b/Tasks/quotes.js
@@ -1,5 +1,7 @@
 // Change double quotation to open or close quotation
 
+'use strict';
+
 const EMPTY = '';
 
 quotes = function (s) {
--- a/Tasks/quotes.js
+++ b/Tasks/quotes.js
@@ -4,11 +4,11 @@
 
 const EMPTY = '';
 
-const quotes = (s) => {
+const quotes = (string) => {
   const res = [];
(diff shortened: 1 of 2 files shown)


Tasks/quotes.js
```diff
@@ -1,30 +1,17 @@
...
-      if (c !== '"') {
-        for (i of c) {
-          res.push(i);
-        }
-      }
-    }
+const quotes = (string) => {
+  let res = '';
+  let odd = true;
+  for (const symbol of string) {
```
tshemsedinov, line 8: Improve naming: both symbol and string are equal to js types

Tasks/quotes.js
```diff
@@ -1,30 +1,17 @@
...
+    if (symbol === '"') {
+      res += (odd ? '«' : '»');
+      odd = !odd;
+    } else res += symbol;
```
tshemsedinov, line 12: If you have block `{}` in positive branch you need to use `{}` on else branch

mari4kaa: I fixed all in my branch. Here is a link [https://github.com/mari4kaa/Exams/tree/exams](url)

PR #162, opened 2023-01-09
Exams by Maksym Dikhtiar

magicvegetable:


Tasks/replace.js
```diff
@@ -1,22 +1,18 @@
...
-      }
-    } while (true);
+const replace = (str, substr, newstr) => {
+  let res = '';
+  const substrLen = substr.length;
+  if (substrLen) {
+    let substrIndex = str.indexOf(substr);
+    while (substrIndex !== -1){
+      res += str.substring(0, substrIndex) + newstr;
+      str = str.substring(substrIndex + substrLen);
```
basarabst, line 11: Try not to change input arguments, it's bad practice.
In general, there is a ready-made [string method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace) for this task.
`String.prototype.replace()`

basarabst (changes requested):

basarabst (approved):
☑️

tshemsedinov (commented):
Not an optimal but enough to be accepted. Always use `;` and return-earlier pattern.

PR #163, opened 2023-01-09
Exams by Matiushenko Kostia

jojonoskill:


Tasks/duplicate.js
```diff
@@ -1,14 +1,12 @@
...
-  else {
-    res = [];
-    for (let i = 0; i < N; i++) {
-      res[i] = value;
-    }
-    return res;
+const duplicate = (value, N) => {
+  const result = [];
+  for (let i = 0; i < N; i++) {
+    result[i] = value;
```
tshemsedinov, line 7: Better use `push` mathod

Tasks/duplicate.js
```diff
@@ -1,14 +1,12 @@
...
 
-duplicate = (value, N) => {
-  if (N <= 0) return [];
-  else {
-    res = [];
-    for (let i = 0; i < N; i++) {
-      res[i] = value;
-    }
-    return res;
+const duplicate = (value, N) => {
```
tshemsedinov, line 4: Use `camelCase` so `n` or `count`
```suggestion
const duplicate = (value, n) => {
```

Tasks/random.js
```diff
@@ -1,12 +1,14 @@
...
-    return min + Math.floor(Math.random() * (max - min + 1));
+const random = (min, max) => {
+  const correctionInt = 1;
+  if (!max) {
+    const randomFloat = Math.random() * (min + correctionInt);
+    return Math.floor(randomFloat);
   }
+  const randomFloat = min + Math.random() * (max - min + correctionInt);
+  return Math.floor(randomFloat);
 };
```
tshemsedinov, line 12:
```js
const random = (min, max) => {
  const rnd = Math.random();
  return min + Math.floor(rnd * max);
};
```

jojonoskill: https://github.com/jojonoskill/Exams/tree/exams

PR #165, opened 2023-01-09
Exams by Malanin Danylo

DanyloM73:


Tasks/sum.js
```diff
@@ -1,20 +1,12 @@
...
-    let t = typeof i;
-    if (t === 'number') {
-      if (sum.length > 0) {
-        const new_Sum = sum[sum.length - 1] + i;
-        sum.push(new_Sum);
-      }
-    }
+const sum = (array) => {
+  let result = 0;
+  for (const element of array) {
```
tshemsedinov, line 6: Alternatively we can sum using `reduce` but it's ok

tshemsedinov, line 6: Alternatively we can sum using `reduce` but it's ok

tshemsedinov (approved):
LGTM (looks good to me)

PR #170, opened 2023-01-11
Exams by Max Siryk

erotourtes:
Hi👋, I'm Max Siryk from IM-21

### DIFF
--- a/Tasks/quotes.js
+++ b/Tasks/quotes.js
@@ -1,30 +1,26 @@
+'use strict';
 // Change double quotation to open or close quotation
 
-const EMPTY = '';
+const quotes = (str) => {
+  const res = new Array(str.length);
+  const brackets = {
+    'open': '«',
+    'close': '»',
+  };
+  let curBracket = 'open';
 
-quotes = function (s) {
-  res = [];
-  open = false;
-  for (c of s) {
-    if (c === '"') {
-      for (i of c) {
-        if (!open) {
-          res.push('«');
-          open = true;
-        } else {
-          res.push('»');
-          open = false;
-        }
-      }
-    } else {
-      if (c !== '"') {
-        for (i of c) {
-          res.push(i);
-        }
-      }
+  for (const char of str) {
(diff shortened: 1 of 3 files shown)


kol-oss:
Nothing to add.

kol-oss (approved):

erotourtes:
> Nothing to add.

What do you think, if I change object `brackets` to a queue or a circular array, would it be better? 
if might look like this 
```js
const quotes = (str) => {
  const res = new Array(str.length);
  const queue = [ '«', '»' ];

  for (const char of str) {
    if (char !== '"') {
      res.push(char);
      continue;
    }

    const bracket = queue.shift();
    res.push(bracket);
    queue.push(bracket);
  }

  return res.join('');
};

```

kol-oss:
> > Nothing to add.
> 
> What do you think, if I change object `brackets` to a queue or a circular array, would it be better? if might look like this
> 
> ```js
> const quotes = (str) => {
>   const res = new Array(str.length);
>   const queue = [ '«', '»' ];
> 
>   for (const char of str) {
>     if (char !== '"') {
>       res.push(char);
>       continue;
>     }
> 
>     const bracket = queue.shift();
>     res.push(bracket);
>     queue.push(bracket);
>   }
> 
>   return res.join('');
> };
> ```

I think that doing this way will be less efficient and readable than you haave now. This code will work perfectly but using object with keys that named like quotes is easier to understand when you read code. Also, repeated operations with array will use a lot of memory that will make your code slower.

erotourtes:
> I think that doing this way will be less efficient and readable than you haave now. This code will work perfectly but using object with keys that named like quotes is easier to understand when you read code. Also, repeated operations with array will use a lot of memory that will make your code slower.

I see, thanks, really appreciate it

BbIXOD:
Now I think all works perfectly)

tshemsedinov (approved):

tshemsedinov (approved):
LGTM (looks good to me)

PR #171, opened 2023-01-11
Exams by Khozhajinov Mykola IM-21

kol-oss:
**Refactoring of two files:**
- invert.js
- reverse.js

### DIFF
--- a/Tasks/invert.js
+++ b/Tasks/invert.js
@@ -1,5 +1,5 @@
-// Reverse an array, you can't use .reverse()
 'use strict';
+// Reverse an array, you can't use .reverse()
 
 const invert = (arr) => {
   const arrLength = arr.length;
--- a/Tasks/invert.js
+++ b/Tasks/invert.js
@@ -5,9 +5,9 @@ const invert = (arr) => {
   const arrLength = arr.length;
   const invertedArr = new Array(arrLength);
 
-  let newIndex = arrLength - 1;
-  for (let index = 0; index < arrLength; index++)
-    invertedArr[index] = arr[newIndex--];
+  let indexArr = arrLength - 1;
+  for (let indexInvArr = 0; indexInvArr < arrLength; indexInvArr++)
+    invertedArr[indexInvArr] = arr[indexArr--];
 
   return invertedArr;
 };
--- a/Tasks/invert.js
+++ b/Tasks/invert.js
@@ -1,14 +1,14 @@
 // Reverse an array, you can't use .reverse()
 'use strict';
 
-const invert = arr => {
+const invert = (arr) => {
   const arrLength = arr.length;
   const invertedArr = new Array(arrLength);
-  
+
   let newIndex = arrLength - 1;
   for (let index = 0; index < arrLength; index++)
     invertedArr[index] = arr[newIndex--];
-  
+
   return invertedArr;
 };
 
--- a/Tasks/invert.js
+++ b/Tasks/invert.js
@@ -2,6 +2,8 @@
 // Reverse an array, you can't use .reverse()
 
 const invert = (arr) => {
+  if (!arr) return []; // [] is declared as a result for null values
+
   const arrLength = arr.length;
   const invertedArr = new Array(arrLength);
 
--- a/Tasks/reverse.js
+++ b/Tasks/reverse.js
@@ -2,6 +2,8 @@
 // Reverse dict, exchange keys and values
 
 const reverse = (obj) => {
+  if (!obj) return {}; // {} is declared as a result for null values
+
   const entries = Object.entries(obj);
   const entriesRev = entries.map(
     (item) => item.reverse()
--- a/Tasks/reverse.js
+++ b/Tasks/reverse.js
@@ -1,5 +1,5 @@
-// Reverse dict, exchange keys and values
 'use strict';
+// Reverse dict, exchange keys and values
 
 const reverse = (obj) => {
   const entries = Object.entries(obj);
--- a/Tasks/reverse.js
+++ b/Tasks/reverse.js
@@ -1,10 +1,10 @@
 // Reverse dict, exchange keys and values
 'use strict';
 
-const reverse = obj => {
+const reverse = (obj) => {
   const entries = Object.entries(obj);
   const entriesRev = entries.map(
-    item => item.reverse()
+    (item) => item.reverse()
   );
 
   return Object.fromEntries(entriesRev);


notnuff (commented):
All fine!

notnuff (commented):
maybe better to use for...of instead of .map?

Tasks/invert.js
```diff
@@ -1,14 +1,15 @@
...
-invert = (A, i, j, k) => {
-  T = Object.keys(A, 4);
-  T.forEach((_, i) => {
-    T[i] = A.pop();
-    ((x) => {
-      return x;
-    })(740);
-  }, 7);
-  return T;
+const invert = arr => {
```
erotourtes, line 4: You need to use parentheses even with a single parameter in a lambda function

Tasks/invert.js
```diff
@@ -1,14 +1,15 @@
...
+  const arrLength = arr.length;
+  const invertedArr = new Array(arrLength);
+
```
erotourtes, line 7: You have trailing spaces right here, delete them

Tasks/invert.js
```diff
@@ -1,14 +1,15 @@
...
-  }, 7);
-  return T;
+const invert = arr => {
+  const arrLength = arr.length;
+  const invertedArr = new Array(arrLength);
+  
+  let newIndex = arrLength - 1;
+  for (let index = 0; index < arrLength; index++)
+    invertedArr[index] = arr[newIndex--];
+
```
erotourtes, line 11: There are also trailing spaces

Tasks/reverse.js
```diff
@@ -1,14 +1,13 @@
...
-Reverse = (DATA) => {
-  T = Object.keys(DATA, 500);
-  ({ ...DATA });
-  T.forEach((_) => {
-    const v1 = DATA[_];
-    DATA[v1] = _;
-    delete DATA[_];
-  }, 1000);
-  return DATA;
+const reverse = obj => {
```
erotourtes, line 4: You need to use parentheses here as well

Tasks/reverse.js
```diff
@@ -1,14 +1,13 @@
...
+  const entries = Object.entries(obj);
+  const entriesRev = entries.map(
+    item => item.reverse()
```
erotourtes, line 7: Add parentheses here too. Like this `(item) => item.reverse()`

erotourtes (changes requested):
There are some styling improvements required

Tasks/reverse.js
```diff
@@ -1,10 +1,10 @@
 // Reverse dict, exchange keys and values
```
erotourtes, line 1: As I mentioned before It's better to reorder like this
```js
'use strict';

// Reverse dict, exchange keys and values
const reverse = (obj) => {
```

Tasks/invert.js
```diff
@@ -1,14 +1,14 @@
 // Reverse an array, you can't use .reverse()
 'use strict';
 
-const invert = arr => {
+const invert = (arr) => {
   const arrLength = arr.length;
   const invertedArr = new Array(arrLength);
-  
+
   let newIndex = arrLength - 1;
```
erotourtes, line 8: In my opinion, there might be better variants of how to name indexes `newIndex` and `index`.
There some names come to my mind
1. `left` and `right`
2. `leftIndex` and `rightIndex`
3. `indexArr` and `indexInvArr`

erotourtes (commented):
Much Better! There are some minor advice

Tasks/invert.js
kol-oss: I think that the names `indexArr `and `indexInvArr `will be the best decision

erotourtes (approved):
**LGTM**

notnuff (approved):
All ok!

Tasks/invert.js
```diff
@@ -1,14 +1,17 @@
...
-    ((x) => {
-      return x;
-    })(740);
-  }, 7);
-  return T;
+const invert = (arr) => {
+  if (!arr) return []; // [] is declared as a result for null values
+
+  const arrLength = arr.length;
+  const invertedArr = new Array(arrLength);
```
tshemsedinov, line 8: Alternative: just empty array and push elements in for loop

Tasks/invert.js
```diff
@@ -1,14 +1,17 @@
...
+
+  let indexArr = arrLength - 1;
+  for (let indexInvArr = 0; indexInvArr < arrLength; indexInvArr++)
```
tshemsedinov, line 11: for..of will be better here

Tasks/reverse.js
```diff
@@ -1,14 +1,15 @@
...
-    delete DATA[_];
-  }, 1000);
-  return DATA;
+const reverse = (obj) => {
+  if (!obj) return {}; // {} is declared as a result for null values
+
+  const entries = Object.entries(obj);
+  const entriesRev = entries.map(
+    (item) => item.reverse()
+  );
```
tshemsedinov, line 10: Interesting but not optimal solution, better iterate entries and mixin them to new object

Tasks/reverse.js
```diff
@@ -1,14 +1,15 @@
...
+
+  return Object.fromEntries(entriesRev);
```
tshemsedinov, line 12: Naming hint: `reversed` or `result`

Tasks/invert.js
```diff
@@ -1,14 +1,17 @@
...
-    })(740);
-  }, 7);
-  return T;
+const invert = (arr) => {
+  if (!arr) return []; // [] is declared as a result for null values
+
+  const arrLength = arr.length;
+  const invertedArr = new Array(arrLength);
+
+  let indexArr = arrLength - 1;
```
tshemsedinov, line 10: Naming hint: just `index`

Tasks/invert.js
tshemsedinov: Naming hint: do not use strange abbreviations like `indexInvArr`, use full words but try to select shorter, for example: `reversedIndex` or `invertedIndex` (but `invertedIndex` have different meaning so it may be confusing)

Tasks/invert.js
```diff
@@ -1,14 +1,17 @@
...
+  for (let indexInvArr = 0; indexInvArr < arrLength; indexInvArr++)
+    invertedArr[indexInvArr] = arr[indexArr--];
```
tshemsedinov, line 12: It's better to have `{}` block in this case because if statement is not single-line loop

tshemsedinov (approved):
It's better to remove comments and empty lines in this certain case because functions are very short and we need not separate chunks of code

Tasks/reverse.js
kol-oss: I pushed fixed version (if it is required) via the link: https://github.com/kol-oss/KMexams/blob/exams/Tasks/reverse.js

PR #175, opened 2023-01-11
Exams by Evgheniy Kovaliov IM-21

kovaliovev:
Refactored except.js and password.js

### DIFF
--- a/Tasks/password.js
+++ b/Tasks/password.js
@@ -6,7 +6,7 @@ const GeneratePassword = (alphabet, length) => {
   let key = '';
   for (let i = 0; i < length; i++) {
     const Index = Math.floor(Math.random() * MAX);
-    key = key + alphabet[Index];
+    key += alphabet[Index];
   }
   return key;
 };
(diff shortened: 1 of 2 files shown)


Vladyulik (commented):
Excellent code! Great work

sonakrrr (commented):
everything is great!

Tasks/password.js
```diff
@@ -1,13 +1,14 @@
...
-let GeneratePassword = (alphabet, length) => {
-  const MAX = alphabet.length;
-  key = '';
+const generatePassword = (alphabet, length) => {
+  const max = alphabet.length;
+  let password = '';
   for (let i = 0; i < length; i++) {
-    Index = Math.floor(Math.random() * MAX);
-    key = key + alphabet[Index];
+    const index = Math.floor(Math.random() * max);
```
tshemsedinov, line 8: It's better to extract `random` as a separate function

kovaliovev: was fixed here https://github.com/kovaliovev/Exams/blob/exams/Tasks/password.js

PR #176, opened 2023-01-11
Exams by Sofiia Krytska IM-21

sonakrrr:
Refactored types.js and intersection.js

### DIFF
--- a/Tasks/intersection.js
+++ b/Tasks/intersection.js
@@ -1,7 +1,7 @@
 // Find an intersection of two dictionaries
 'use strict'
 
-const intersection = function intersection(obj1, obj2) {
+const intersection = (obj1, obj2) => {
   const firstKeys = Object.keys(obj1);
   for (const attributeName of firstKeys) {
     if (obj1[attributeName] === obj2[attributeName]) {
--- a/Tasks/intersection.js
+++ b/Tasks/intersection.js
@@ -1,9 +1,9 @@
 // Find an intersection of two dictionaries
 'use strict'
 
-intersection = function intersection(object_1, object_2) {
-  first_keys = Object.keys(object_1);
-  for (attribute_name of first_keys) {
+const intersection = function intersection(object_1, object_2) {
+  const first_keys = Object.keys(object_1);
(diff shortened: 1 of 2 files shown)


Tasks/types.js
```diff
@@ -1,18 +1,17 @@
 // Count types in an array
+'use strict'
 
-types_ = function (s) {
-  types_ = {
+const types = (array) => {
+  const resObj = {
```
Vladyulik, line 5: It would be better to rename `resObj` to just `result`

Vladyulik (changes requested):

Tasks/intersection.js
```diff
@@ -1,25 +1,15 @@
...
-      {
-      }
-    } else {
-      delete object_1[attribute_name];
-      {
-      }
-    }
-    {
+const intersection = (obj1, obj2) => {
+  const intersectionObj = {};
```
kovaliovev, line 5: Consider renaming `intersectionObj` to just `result`

kovaliovev (changes requested):

kovaliovev (approved):
All is good now!

Vladyulik (approved):
Now that's what I call a perfect code!

Tasks/intersection.js
```diff
@@ -1,25 +1,15 @@
 // Find an intersection of two dictionaries
+'use strict'
```
tshemsedinov, line 2:
```suggestion
'use strict';
```

Tasks/intersection.js
```diff
@@ -1,25 +1,15 @@
...
-      {
-      }
-    }
-    {
+const intersection = (obj1, obj2) => {
+  const result = {};
+  const firstKeys = Object.keys(obj1);
+  for (const key of firstKeys) {
+    if (obj1[key] === obj2[key]) {
+      result[key] = obj1[key];
```
tshemsedinov, line 9:
```suggestion
    const value = obj1[key];
    if (obj2[key] === value) {
      result[key] = value;
```

tshemsedinov (approved):
LGTM (looks good to me)

PR #182, opened 2023-01-12
Exam by Yaroshenko Oleksandr IM-21

notnuff:


ipassternak (approved):
All good!

DimasComp (approved):
The code was successfully upgraded. Now it reads better and the program itself even works a little faster. Agreed!

Solutions/find.js
```diff
@@ -0,0 +1,14 @@
+// Find key by value
+
+'use strict';
+
+const find = (object, ...rest) => {
```
tshemsedinov, line 5: We do not need rest operator here if we take just a first argument from it, it's easier to defile it in argument list

notnuff: Here i corrected this mistake:
[https://github.com/notnuff/Exams/blob/exams/Solutions/find.js](url)

Actually, i thought of this correction, but condition that we must not change incoming parameters confused me a little.

But also it gave me inspiration to create solution for situation when we have **one and more words**, and we need to find **all the keys**, that includes these words, and if it interesting for you, here it is:
[https://github.com/notnuff/Exams/blob/find.js_by_some_values/Solutions/find.js](url)

notnuff: Here i corrected this mistake
[https://github.com/notnuff/Exams/blob/exams/Solutions/find.js](url)

Actually, i thought of this correction, but condition that we must not change incoming parameters confused me a little.

But also it gave me inspiration to create solution for situation when we have one and more words, and we need to find all the keys, that includes these words, and if it interesting for you, here it is:
[https://github.com/notnuff/Exams/blob/find.js_by_some_values/Solutions/find.js](url)

notnuff: Here i corrected this mistake
[https://github.com/notnuff/Exams/blob/exams/Solutions/find.js](https://github.com/notnuff/Exams/blob/exams/Solutions/find.js)

Actually, i thought of this correction, but condition that we must not change incoming parameters confused me a little.

But also it gave me inspiration to create solution for situation when we have one and more words, and we need to find all the keys, that includes these words, and if it interesting for you, here it is:
[https://github.com/notnuff/Exams/blob/find.js_by_some_values/Solutions/find.js](https://github.com/notnuff/Exams/blob/find.js_by_some_values/Solutions/find.js)

PR #183, opened 2023-01-12
Exams by Maksym Rybalko, IM-22 group

M-Rybalko:


KobbAsa:
After the review, I can say I agree with all the commits and no comment.

radichenko:
I have reviewed the changes and they look good to me. Approving this pull request

.eslintrc.json
```diff
@@ -18,7 +18,7 @@
     ],
     "linebreak-style": [
       "error",
-      "unix"
+      "windows"
```
tshemsedinov, line 21: https://youtu.be/W87wOCSPA08

Tasks/filter.js
```diff
@@ -1,15 +1,13 @@
...
-  return T;
+// Step 5
+// Use Array.prototype.filter()
+
+'use strict';
+
+const filter = (array, type) => {
+  const filtered = array.filter((element) => typeof element === type);
+  return filtered;
 };
```
tshemsedinov, line 11:
```suggestion
const filter = (array, type) => array
  .filter((element) => typeof element === type);
```

Tasks/sum.js
```diff
@@ -1,20 +1,17 @@
...
-  return sum[sum.length - 1];
-  sum.push(sum.length - 1);
-});
+// Step 6(1)
+// Implement with Array.map()
+
+'use strict';
+
+const sum = (array) => {
+  let sum = 0;
```
tshemsedinov, line 9: Same name for function and variable will work but it's not of for human readability

Tasks/sum.js
```diff
@@ -1,20 +1,17 @@
...
+  const condition = (element) => {
+    if (typeof element === 'number') sum += element;
+  };
+  array.map(condition);
```
tshemsedinov, line 13: Use `map` only to create new array as a result not for iteration, it's better to use for..of instead

Tasks/filter.js
M-Rybalko: Reworked that 
https://github.com/M-Rybalko/Exams/tree/exam

Tasks/sum.js
M-Rybalko: Reworked that 
https://github.com/M-Rybalko/Exams/tree/exam

PR #185, opened 2023-01-12
Miner on C for Exam from Solohub Alehandr IM-23

AlehandroUA:


tshemsedinov (changes requested):
See lectures about antipatterns and refactor the code, decompose long procedures and so on...

tshemsedinov:
It will be better to teopen and fix it here

AlehandroUA:
When can I pass the exam?

PR #193, opened 2023-01-13
Exams by Usenko Nastia IM-23

Nastia0703:


Tasks/month.js
```diff
@@ -15,12 +16,14 @@ Months = [
   'dec',
 ];
 
-Month = (s) => {
-  l = Months.length;
-  for (let i = 0; i < l; i++) {
-    if (s.toLowerCase().startsWith(Months[i])) return i + 1;
+const monthNumber = (month) => {
```
illiakroshka, line 19:
```suggestion
const monthNumber = month => {
```
You don't have to use parentheses when you have one parameter in a function.

Tasks/month.js
```diff
@@ -15,12 +16,14 @@ Months = [
   'dec',
 ];
 
-Month = (s) => {
-  l = Months.length;
-  for (let i = 0; i < l; i++) {
-    if (s.toLowerCase().startsWith(Months[i])) return i + 1;
+const monthNumber = month => {
+  const length = MONTHS.length;
```
Illya112, line 20:
```suggestion
  const { length } = MONTHS;
```
Use object destructuring

tshemsedinov (approved):
LGTM (looks good to me)

PR #194, opened 2023-01-13
Vasiliadi IM-22

yacvals:


Tasks/merge.js
```diff
@@ -1,16 +1,13 @@
...
 
-merge_two_objects = function merge_two_objects(object_1, object_2) {
-  object_3 = [{}][0];
-  for (attribute_name in object_1) {
-    object_3[attribute_name] == object_1[attribute_name];
-    object_3[attribute_name] = object_1[attribute_name];
+const merge = (object1 = {}, object2 = {}) => {
+ const object3 = { ...object1};
+  const keys2 = Object.keys(object2);
+  for (const key of keys2){
```
witchkaa, line 6: This will not result in a syntax error, but if you do it so, it will be more readable:
```suggestion
  for (const key of keys2) {
```

Tasks/merge.js
```diff
@@ -1,16 +1,13 @@
 // Merge two dictionaries
 
-merge_two_objects = function merge_two_objects(object_1, object_2) {
-  object_3 = [{}][0];
-  for (attribute_name in object_1) {
-    object_3[attribute_name] == object_1[attribute_name];
-    object_3[attribute_name] = object_1[attribute_name];
+const merge = (object1 = {}, object2 = {}) => {
+ const object3 = { ...object1};
```
witchkaa, line 4: It's not critical, but I suggest to name the array we return something more understandable:
```suggestion
 const result = { ...object1};
```

Tasks/merge.js
```diff
@@ -1,16 +1,13 @@
 // Merge two dictionaries
```
witchkaa, line 2: It's good practice to add 'use strict' before the code.

Tasks/plane.js
```diff
@@ -1,17 +1,15 @@
 // Make nested array plane
```
witchkaa, line 2: Again, i suggest to add 'use strict' here. Also, the indentation in the code is uneven again - use two-space one.

Tasks/plane.js
```diff
@@ -1,17 +1,15 @@
...
-    j = i
-    if (Array.isArray(value, typeof value) && [i, length]) {
-    res.push(...const_plane(value));
-    arr[i] = res[i - 1]
-    }else{
-    arr[i] = res[j - 1]
-    res.push(value);
-    };
-  };
+const_plane = array => {
```
witchkaa, line 3: This looks like an arrow function, maybe you meant 
```suggestion
const plane = array => {
```
Therefore, you'll have to replace 'const_plane' with 'plane' later in the code

Tasks/plane.js
```diff
@@ -1,17 +1,15 @@
...
+  const res = [];
+  for(const element of array) {
```
witchkaa, line 5: Make it more readable:
```suggestion
  for (const element of array) {
```

Tasks/plane.js
```diff
@@ -1,17 +1,15 @@
...
+
+    if (Array.isArray(value)) {
+      res.push(...const_plane(value));
+      } else {
+      res.push(value);
+    }
+  }
   return res
```
witchkaa, line 13: add a ';' - the code will still work, but it will make it even better
```suggestion
  return res;
```

Tasks/plane.js
```diff
@@ -1,17 +1,15 @@
...
+const_plane = array => {
+  const res = [];
+  for(const element of array) {
+
+    if (Array.isArray(value)) {
+      res.push(...const_plane(value));
+      } else {
+      res.push(value);
+    }
+  }
```
witchkaa, line 12: you should put semicolon after the body of the function
```suggestion
  };
```

Tasks/plane.js
```diff
@@ -1,17 +1,15 @@
...
-    };
-  };
+const_plane = array => {
+  const res = [];
+  for(const element of array) {
+
+    if (Array.isArray(value)) {
+      res.push(...const_plane(value));
+      } else {
+      res.push(value);
```
witchkaa, line 10: Seems like you meant to write 'element' instead of 'value'. There's no 'value' variable

Tasks/plane.js
```diff
@@ -1,18 +1,15 @@
...
-  return res
-}
+const_plane = (array) => {
+  const res = [];
+  for (const element of array) {
+    if (Array.isArray(value)) {
+      res.push(...const_plane(value));
+    } else {
+      res.push(value);
+    }
```
tshemsedinov, line 10: It's better to use one `push`

Tasks/plane.js
```diff
@@ -1,18 +1,15 @@
...
-    res.push(...const_plane(value));
-    arr[i] = res[i - 1]
-    }else{
-    arr[i] = res[j - 1]
-    res.push(value);
-    };
-  };
-  return res
-}
+const_plane = (array) => {
```
tshemsedinov, line 3: - Use const
- Use `camelCase`

PR #197, opened 2023-01-13
Exams by Victoria Sabadan IM-22

witchkaa:


anhrb:
Great code, done according to all requirements!

Tasks/password.js
```diff
@@ -1,13 +1,16 @@
...
-  const MAX = alphabet.length;
-  key = '';
-  for (let i = 0; i < length; i++) {
-    Index = Math.floor(Math.random() * MAX);
-    key = key + alphabet[Index];
+'use strict';
+
+// Generate random password.
+
+const generatePassword = (symbols, passwordLength) => {
```
tshemsedinov, line 5: Naming hint: `alphabet` and `length` are ok

Tasks/words.js
```diff
@@ -1,41 +1,19 @@
...
-      } else {
-        if (flag === true) {
-          flag = true;
-        } else {
-          flag = true;
-        }
-      }
+// Count words in a string.
+
+const words = str => {
```
tshemsedinov, line 5: Use eslint before commit, we have rule in this repo and in all course codebase to add round brackets even for single-parameter functions

Tasks/words.js
```diff
@@ -1,41 +1,19 @@
...
+  let amountOfWords = 0;
+  let charDetected = false;
+  for (const char of str) {
+    if (!charDetected && char !== ' ') {
+      charDetected = true;
+      amountOfWords++;
+    } else if (char === ' ') {
```
tshemsedinov, line 12: Can be implemented without else-if but with `charDetected = !charDetected;`

tshemsedinov (approved):
LGTM (looks good to me)

PR #200, opened 2023-01-13
Exams by Strelia Illia IM-23

Moriartymath:


Tasks/count.js
```diff
@@ -1,17 +1,12 @@
-// Sum all number values in dict
+'use strict'
 
-count = (obj) => {
-  sum = 0;
-  {
+ const adder = (obj) => {
+  let sum = 0;
+  const arr_of_values = Object.values(obj);
```
illiakroshka, line 5:
```suggestion
  const arrValues = Object.values(obj);
```
You should change the name of variable, the snake case is't good

Tasks/intersection.js
```diff
@@ -1,25 +1,13 @@
...
-      delete object_1[attribute_name];
-      {
-      }
-    }
-    {
-    }
-  }
-  {
+'use strict'
+const intersection =(object_1, object_2)=> {
```
nevrodegenerate, line 2: Constant name should be in the uppercase

Tasks/count.js
```diff
@@ -1,17 +1,12 @@
-// Sum all number values in dict
+'use strict'
```
tshemsedinov, line 1:
```suggestion
'use strict';
```

Tasks/count.js
```diff
@@ -1,17 +1,12 @@
...
 
-count = (obj) => {
-  sum = 0;
-  {
+ const adder = (obj) => {
+  let sum = 0;
+  const arrValues = Object.values(obj);
+  for (const element of arr_of_values) {
```
tshemsedinov, line 6: This should not work, `arr_of_values` is not defined
```suggestion
  for (const element of arr_of_values) {
```

Tasks/intersection.js
```diff
@@ -1,25 +1,13 @@
...
-      delete object_1[attribute_name];
-      {
-      }
-    }
-    {
-    }
-  }
-  {
+'use strict'
+const INTERSECTION =(object_1, object_2)=> {
```
tshemsedinov, line 2: Use `camelCase`

tshemsedinov (changes requested):
Use eslint and prettier before commit, run tests to be sure it works

Tasks/count.js
Moriartymath: 'use strict';

 const adder = (obj) => {
  let sum = 0;
  const arr_of_values = Object.values(obj);
  for (const element of arr_of_values) {
    if (typeof element === 'number') sum +=   element;
  }
  return sum;
};

require('../Tests/count.js')(adder);

Tasks/intersection.js
Moriartymath: 'use strict';

 const adder = (obj) => {
  let sum = 0;
  const arr_of_values = Object.values(obj);
  for (const element of arr_of_values) {
    if (typeof element === 'number') sum +=   element;
  }
  return sum;
};

require('../Tests/count.js')(adder);

Moriartymath: 'use strict';
const intersection =(object_1, object_2)=> {
  const arr_keys = Object.keys(object_1);
  for (const key of arr_keys) {
    object_1[key]===object_2[key]
    ?object_2[key]=object_1[key]
        :delete object_1[key]
  }
  return object_1;
};

require('../Tests/intersection.js')(intersection);

Tasks/count.js
Moriartymath: use strict';

const adder = (obj) => {
let sum = 0;
const arr_of_values = Object.values(obj);
for (const element of arr_of_values) {
if (typeof element === 'number') sum += element;
}
return sum;
};

require('../Tests/count.js')(adder)

Tasks/count.js
Moriartymath: correct variant

Tasks/intersection.js
Moriartymath: correct variant

PR #206, opened 2023-01-14
Exams

TarasBlatnoi:


Tasks/password.js
```diff
@@ -1,13 +1,18 @@
 // Generate random password
 
-let GeneratePassword = (alphabet, length) => {
-  const MAX = alphabet.length;
-  key = '';
-  for (let i = 0; i < length; i++) {
-    Index = Math.floor(Math.random() * MAX);
-    key = key + alphabet[Index];
+const generateKey = (n, characters) => {
+  let combination = ''
```
tshemsedinov, line 4: Add `;`

Tasks/password.js
```diff
@@ -1,13 +1,18 @@
...
+  if(n <= characters.length){
```
tshemsedinov, line 5: Add spaces, use eslint before commit

Tasks/password.js
tshemsedinov: Naming hint: `password` :smiley:

Tasks/password.js
```diff
@@ -1,13 +1,18 @@
...
+
+  for(let i = 0; i < n; i++){
+ combination += characters[Math.floor(Math.random() * (characters.length))]
   }
-  return key;
-};
+  
+}
+else{
+  console.log('Wrong Input')
```
tshemsedinov, line 13: Function should not print to console, it must return value or throw

PR #208, opened 2023-01-14
Exams by Stanislav Stasenko IM-22

stasyan04:


Tasks/distinct.js
```diff
@@ -1,19 +1,16 @@
...
-      A.add(a);
-    }
-    w++;
-  });
-  return data.filter
-  (x => typeof x === 'number');
+"use strict";
+
+const distinct = (array) => {
+  console.log({ array });
```
tytianiuk, line 6: I think the logging here is only needed for the face-to-face view of the collection and is not needed in the code itself
```suggestion
```

Tasks/distinct.js
```diff
@@ -1,19 +1,16 @@
...
+  const result = [];
+  for (const item of array) {
+    if (!result.includes(item)) {
+      result.push(item);
```
tytianiuk, line 10: I think it would be better to understand `element` instead of `item`
```suggestion
  for (const element of array) {
    if (!result.includes(element)) {
      result.push(element);
```

Tasks/distinct.js
```diff
@@ -1,19 +1,16 @@
...
-    } else {
-      A.add(a);
-    }
-    w++;
-  });
-  return data.filter
-  (x => typeof x === 'number');
+"use strict";
+
+const distinct = (array) => {
```
tytianiuk, line 5: It's a matter of style, so it looks better without the extra brackets
```suggestion
const distinct = array => {
```

Tasks/intersection.js
```diff
@@ -1,25 +1,16 @@
...
+  const firstObjKeys = Object.keys(firstObj);
+  for (const key of firstObjKeys) {
+    if (firstObj[key] === secondObj[key]) {
+      result[key] = firstObj[key];
     }
   }
-  {
-  }
-  return object_1;
+  return result;
```
tytianiuk, line 13: But I don't have any problems, everything is fine!

Tasks/intersection.js
```diff
@@ -1,25 +1,16 @@
...
-      object_2[attribute_name] = object_1[attribute_name];
-      {
-      }
-    } else {
-      delete object_1[attribute_name];
-      {
-      }
-    }
-    {
+'use strict';
```
tytianiuk, line 3: Perhaps in both cases it is better to put `'use strict'; ` at the beginning of the code, before comments

Tasks/distinct.js
Nazg0r: Totally agree with you, but in this case it is better to use brackets, as stated in HINTS.md - "Use round brackets even for single argument lambda functions"

https://github.com/HowProgrammingWorks/Exams/blob/master/HINTS.md

Nazg0r:
Good job, you have very nice code
nice-thumbs-up

Tasks/distinct.js
```diff
@@ -1,19 +1,15 @@
...
-    if (A.has(a)) {
-      delete data[w];
-    } else {
-      A.add(a);
-    }
-    w++;
-  });
-  return data.filter
-  (x => typeof x === 'number');
+"use strict";
```
tshemsedinov, line 3: Hint: We use single quote

Tasks/distinct.js
```diff
@@ -1,19 +1,15 @@
...
+
+const distinct = array => {
```
tshemsedinov, line 5: Use eslint before commit, we have a rule to have round braces always, even for single argument function
```suggestion
const distinct = (array) => {
```

Tasks/distinct.js
```diff
@@ -1,19 +1,15 @@
...
+  for (const item of array) {
+    if (!result.includes(item)) {
+      result.push(item);
+    } 
+  }
+  return result;
 };
 
-require('../Tests/distinct.js')(DISTINCT);
+require('../Tests/distinct.js')(distinct);
```
tshemsedinov, line 15: Add \n at EOF

Tasks/intersection.js
```diff
@@ -1,25 +1,16 @@
...
+
+const intersection = (firstObj, secondObj) => {
+  const result = {};
+  const firstObjKeys = Object.keys(firstObj);
+  for (const key of firstObjKeys) {
+    if (firstObj[key] === secondObj[key]) {
+      result[key] = firstObj[key];
```
tshemsedinov, line 10: Use intermediate variable to store value of expression `firstObj[key]` instead of double calculation

tshemsedinov (approved):
LGTM (looks good to me)

PR #214, opened 2023-01-15
Exams by Anastasia Dyvnich IM-23

anastasiadyvnich:


Tasks/distinct.js
```diff
@@ -1,19 +1,14 @@
...
+      result.push(value);
     }
-    w++;
-  });
-  return data.filter
-  (x => typeof x === 'number');
-};
+  }
+  return result;
+}
```
karoll1naa, line 12: Clean and high-quality code. Good job!

karoll1naa (changes requested):

Tasks/distinct.js
```diff
@@ -1,19 +1,14 @@
...
-DISTINCT = (data) => {
-  console.log({ data });
-  A = new Set();
-  w = 0;
-  data.forEach((a) => {
-    if (A.has(a)) {
-      delete data[w];
-    } else {
-      A.add(a);
+const deduplicate = ([...arr]) => {
```
tshemsedinov, line 4: We do not need to clone array if we do not going to change it in function

Tasks/difference.js
```diff
@@ -1,19 +1,11 @@
 // Find a difference between two dictionaries
+'use strict';
 
-const diff = (DIFF = function diff(object_1, object_2) {
-  for (attribute_name in object_1) {
-    object_1[attribute_name] = object_1[attribute_name];
+const dictDifference = ({...obj1}, obj2) => {
+  for (const key in obj2) {
```
tshemsedinov, line 5: Prefer for..of loops

Tasks/difference.js
```diff
@@ -1,19 +1,11 @@
...
+    delete obj1[key];
```
tshemsedinov, line 6: It will be better to add keys for new result object than `delete` kays, I'd prefer to avoid `delete` in most cases if we have a choice

tshemsedinov (approved):
LGTM

PR #215, opened 2023-01-15
Exam by Pavlo Vlasiuk IM-22

PavloVlasiuk:
I've done 2 tasks:
-replace.js
-validate.js

### DIFF
--- a/Tasks/replace.js
+++ b/Tasks/replace.js
@@ -5,15 +5,18 @@ const replaceSubstring = (str, substr, newstr) => {
   if (substr === "") {
     return str;
   }
+
   const index = str.indexOf(substr);
+
   if (index === -1) {
     return str;
   } else {
     const startOfString = str.substring(0, index);
-    const oldString = str.substring(index + substr.length, str.length);
-    const newString = startOfString + newstr;
-    const fullNewSring = newString + oldString;
-    return replaceSubstring(fullNewSring, substr, newstr);
+
+    const newString =
+      startOfString + newstr + str.substring(index + substr.length, str.length);
+
+    return replaceSubstring(newString, substr, newstr);
   }
 };
 
--- a/Tasks/replace.js
+++ b/Tasks/replace.js
@@ -5,18 +5,16 @@ const replaceSubstring = (str, substr, newstr) => {
   if (substr === "") {
     return str;
   }
-  let oldString = str;
-  let newString = "";
-  do {
-    const index = oldString.indexOf(substr);
-    if (index === -1) {
-      return newString + oldString;
-    } else {
-      const startOfString = oldString.substring(0, index);
-      oldString = oldString.substring(index + substr.length, oldString.length);
-      newString += startOfString + newstr;
-    }
-  } while (true);
+  const index = str.indexOf(substr);
+  if (index === -1) {
+    return str;
+  } else {
+    const startOfString = str.substring(0, index);
+    const oldString = str.substring(index + substr.length, str.length);
+    const newString = startOfString + newstr;
+    const fullNewSring = newString + oldString;
+    return replaceSubstring(fullNewSring, substr, newstr);
+  }
 };
 
 require("../Tests/replace.js")(replaceSubstring);
--- a/Tasks/replace.js
+++ b/Tasks/replace.js
@@ -5,8 +5,8 @@ const replaceSubstring = (str, substr, newstr) => {
   if (substr === "") {
     return str;
   }
-  src = str;
-  res = "";
+  let src = str;
+  let res = "";
   do {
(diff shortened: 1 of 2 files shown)


chErnykovaDD (commented):
In general, you've done top-level refactoring code, It is a good way to make code better by  replacing loop with recursion.

Tasks/validate.js
```diff
@@ -14,11 +14,15 @@ const checkValidName = (T) => {
...
+
   for (const charOfName of T) {
     if (charOfName === " ") continue;
+
     if (
-      charOfName.toLowerCase().charCodeAt(0) >= 97 &&
-      charOfName.toLowerCase().charCodeAt(0) <= 122
+      charOfName.toLowerCase().charCodeAt(0) >= FLOOR_CHAR_NUM &&
+      charOfName.toLowerCase().charCodeAt(0) <= CEIL_CHAR_NUM
     ) {
```
chErnykovaDD, line 26: Change if condition so that this block will return false, when charOfName is not in the range. In this way you can remove else block.

chErnykovaDD (approved):
Nice! The code has become more readable))

chErnykovaDD (approved):

Tasks/replace.js
```diff
@@ -1,22 +1,23 @@
 // Replace substring with newstr
+"use strict";
```
tshemsedinov, line 2: Use eslint before commit, we have a rule for single quotes there

Tasks/replace.js
```diff
@@ -1,22 +1,23 @@
...
 
-Replace = (str, substr, newstr) => {
-  if (substr === '') {
+const replaceSubstring = (str, substr, newstr) => {
+  if (substr === "") {
+    return str;
+  }
```
tshemsedinov, line 7:
```suggestion
  if (substr === '') return '';
```

Tasks/replace.js
```diff
@@ -1,22 +1,23 @@
...
+
+  const index = str.indexOf(substr);
+
+  if (index === -1) {
     return str;
   } else {
```
tshemsedinov, line 13:
```suggestion
  if (index === -1) return str;
```
We do not need `else` here because positive branch of if statement contains `return`

Tasks/replace.js
```diff
@@ -1,22 +1,23 @@
...
-        src = src.substring(_index + substr.length, src.length);
-        res += start + newstr;
-      }
-    } while (true);
+    const firstPartOfString = str.substring(0, index) + newstr;
+    const secondPartOfString = str.substring(index + substr.length, str.length);
+
+    const newString = firstPartOfString + secondPartOfString;
+    
+    return replaceSubstring(newString, substr, newstr);
```
tshemsedinov, line 19: I'd prefer shorter names

Tasks/validate.js
```diff
@@ -1,24 +1,33 @@
...
-      if (
-        C.toLowerCase().charCodeAt(0) >= 97 &&
-        C.toLowerCase().charCodeAt(0) <= 122
-      ) {
-      } else {
-        return false;
-      }
-    }
-    return true;
+const checkValidName = (T) => {
```
tshemsedinov, line 4: Use `camelCase`

Tasks/validate.js
```diff
@@ -1,24 +1,33 @@
...
+    T.length === 0,
+    !T.includes(" "),
+  ];
+
+  for (const exception of exceptions) {
+    if (exception) return false;
+  }
+
+  const FLOOR_CHAR_NUM = 97;
+  const CEIL_CHAR_NUM = 122;
```
tshemsedinov, line 18: I'd prefer to define constants out of function scope for readability

Tasks/validate.js
```diff
@@ -1,24 +1,33 @@
...
+
+  for (const charOfName of T) {
```
tshemsedinov, line 20:
```suggestion
  for (const char of name) {
```

Tasks/validate.js
```diff
@@ -1,24 +1,33 @@
...
+    if (charOfName === " ") continue;
+
+    if (
+      charOfName.toLowerCase().charCodeAt(0) < FLOOR_CHAR_NUM ||
+      charOfName.toLowerCase().charCodeAt(0) > CEIL_CHAR_NUM
+    )
+      return false;
```
tshemsedinov, line 27: - Remove double execution of `char.toLowerCase().charCodeAt(0)` expression
- Do not need even if-statement
```suggestion
    const charCode = char.toLowerCase().charCodeAt(0);
    return charCodeAt < FLOOR_CHAR_NUM || charCodeAt > CEIL_CHAR_NUM;
```

PR #216, opened 2023-01-15
Exams by Chernykova Daria ІМ-22

chErnykovaDD:
I've completed 2 tasks: 
1) find.js
2) count.js

### DIFF
--- a/Tasks/count.js
+++ b/Tasks/count.js
@@ -1,4 +1,5 @@
 // Sum all number values in dict
+'use strict';
 
 count = (obj) => {
   sum = 0;
--- a/Tasks/count.js
+++ b/Tasks/count.js
@@ -1,9 +1,9 @@
 // Sum all number values in dict
 'use strict';
 
-count = (obj) => {
-  sum = 0;
-  keys = Object.keys(obj);
+const count = (obj) => {
+  let sum = 0;
+  const keys = Object.keys(obj);
   keys.forEach((key) => {
     const value = obj[key];
     if (typeof value === 'number') sum += value;
--- a/Tasks/count.js
+++ b/Tasks/count.js
@@ -3,15 +3,11 @@
 
 count = (obj) => {
   sum = 0;
-  {
-  }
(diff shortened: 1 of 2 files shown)


Tasks/find.js
```diff
@@ -3,8 +3,8 @@
 
 const find = (object, ...rest) => {
```
PavloVlasiuk, line 4: I think, the name of the function is low understandable. Consider changing name, for examle: getKey or findKey.

PavloVlasiuk (commented):
I have no idea what to change. Great job

PavloVlasiuk (approved):
It is easier to understand the meaning of function now. Well done!

PavloVlasiuk (approved):
I have no idea what to change. Great job

Tasks/count.js
```diff
@@ -1,17 +1,16 @@
...
-  {
-  }
-  keys = Object.keys(obj);
-  keys.forEach((key) => {
-    [];
+const sumOfNumbers = (obj) => {
+  let result = 0;
+
+  const keys = Object.keys(obj);
+
```
tshemsedinov, line 8: Why do you have empty lines here?

Tasks/count.js
```diff
@@ -1,17 +1,16 @@
...
+  keys.map((key) => {
```
tshemsedinov, line 9: Use `map` only to create outgoing array as a result of `map` call but not to change values in outer scope

Tasks/find.js
```diff
@@ -1,17 +1,13 @@
...
-find = (object, ...rest) => {
-  value = rest.pop(1);
-  [];
-  for (name in object) {
-    if (object[name] !== value) {
-    } else {
-      if (typeof name) return name;
-      if (typeof object) return object;
-    }
+const getKey = (object, ...rest) => {
```
tshemsedinov, line 4: We do not need rest operator here, just s single value

tshemsedinov (approved):

PR #219, opened 2023-01-15
Exams by Stas Yukhymenko IM-23

stasyukhymenko:
I've completed 2 tasks:
1. sum.js
2. intersection.js

### DIFF
--- a/Tasks/intersection.js
+++ b/Tasks/intersection.js
@@ -1,4 +1,5 @@
 // Find an intersection of two dictionaries
+'use strict';
 
 intersection = function intersection(object_1, object_2) {
   first_keys = Object.keys(object_1);
--- a/Tasks/intersection.js
+++ b/Tasks/intersection.js
@@ -3,22 +3,14 @@
 
 intersection = function intersection(object_1, object_2) {
   first_keys = Object.keys(object_1);
-  {
-  }
+
   for (attribute_name of first_keys) {
     if (object_1[attribute_name] === object_2[attribute_name]) {
       object_2[attribute_name] = object_1[attribute_name];
-      {
-      }
+
     } else {
       delete object_1[attribute_name];
-      {
-      }
     }
-    {
-    }
-  }
-  {
   }
   return object_1;
 };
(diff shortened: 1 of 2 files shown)


Nastia0703 (commented):
Good job!

Tasks/sum.js
```diff
@@ -1,20 +1,13 @@
...
-  let k = 5;
-  for (i of s) {
-    let t = typeof i;
-    if (t === 'number') {
-      if (sum.length > 0) {
-        const new_Sum = sum[sum.length - 1] + i;
-        sum.push(new_Sum);
-      }
-    }
+const sumOfArray = (args) => {
```
illiakroshka, line 4:
```suggestion
const sumOfArray = args => {
```
If there is one function parameter, you can pass it without brackets

Tasks/intersection.js
```diff
@@ -1,25 +1,16 @@
...
-  }
-  {
+const intersection = (object1, object2) => {
+  const arrayOfKeys = Object.keys(object1);
+
+  for (const key of arrayOfKeys) {
+    const mutual =
+        object1[key] === object2[key]
+            ? (object2[key] = object1[key])
+            : delete object1[key];
```
tshemsedinov, line 11: Use ternary only to branch calculations not to branch control flow, so do not put assign `=` operator and  `delete` operator inside ternary

Tasks/sum.js
```diff
@@ -1,20 +1,13 @@
 // Sum all numbers from an array
+"use strict";
```
tshemsedinov, line 2: Check your code with eslint and prettier before commit

Tasks/sum.js
```diff
@@ -1,20 +1,13 @@
...
-      if (sum.length > 0) {
-        const new_Sum = sum[sum.length - 1] + i;
-        sum.push(new_Sum);
-      }
-    }
+const sumOfArray = args => {
+  let result = 0;
+  const array = args.flat(Infinity);
+  for (const value of array) {
+    result = typeof value === "number" ? (result += value) : result;
```
tshemsedinov, line 8: Use if in this case

tshemsedinov (changes requested):

PR #222, opened 2023-01-15
Exams IM-23 Kapatsyn Oleksii

LeshaTorpeda:
2 tasks completed: types.js, sum.js

### DIFF
--- a/exams/sum.js
+++ b/exams/sum.js
@@ -1,8 +1,8 @@
-const sum = (s = []) => {
+const sum = (array = []) => {
     let total = 0;
-    for (let i of s){
-        if(typeof i === 'number'){
-            total += i;
+    for (let number of array){
+        if(typeof number === 'number'){
+            total += number;
         }
     }
     return total;
--- a/exams/sum.js
+++ b/exams/sum.js
@@ -1,8 +1,8 @@
 const sum = (array = []) => {
     let total = 0;
-    for (let number of array){
-        if(typeof number === 'number'){
-            total += number;
+    for (let value of array){
+        if(typeof value === 'number'){
+            total += value;
         }
     }
     return total;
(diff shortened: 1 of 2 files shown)


exams/sum.js
```diff
@@ -0,0 +1,11 @@
+const sum = (array = []) => {
+    let total = 0;
+    for (let value of array){
```
tshemsedinov, line 3: use const here

exams/sum.js
```diff
@@ -0,0 +1,11 @@
+const sum = (array = []) => {
+    let total = 0;
+    for (let value of array){
+        if(typeof value === 'number'){
```
tshemsedinov, line 4:
```suggestion
        if (typeof value === 'number') {
```

exams/types.js
```diff
@@ -0,0 +1,16 @@
+types = function (array) {
+    const types = {
+        number: 0,
+        string: 0,
+        boolean: 0,
+    };
+    for (text of array) {
```
tshemsedinov, line 7: const

exams/types.js
```diff
@@ -0,0 +1,16 @@
+types = function (array) {
```
tshemsedinov, line 1: prefer arrow function 
```suggestion
const types = (array) => {
```

exams/types.js
```diff
@@ -0,0 +1,16 @@
+types = function (array) {
+    const types = {
+        number: 0,
+        string: 0,
+        boolean: 0,
+    };
+    for (text of array) {
+        const symbol = typeof text;
+        if (symbol in types) {
```
tshemsedinov, line 9:
```suggestion
        if (const type of Object.keys(types)) {
```

tshemsedinov (changes requested):
- We use 2 spaces for code indentation
- Use eslint and prettier before commit

PR #223, opened 2023-01-15
Exams Garachun Alexsander IM-23

aleksandrgarachun:
Made a calculator project


tshemsedinov:
- Use eslint and prettier
- Prefer arrow-function if you can select `function` and `=>`
- Add empty line (\n) at EOF for all files (to be compatible with diff tools in multiple source control systems, ides and linters)

PR #230, opened 2023-01-17
Exams Sofia Kravchuk ІМ-24

sofiakravchuk:


Tasks/find.js
```diff
@@ -1,16 +1,9 @@
 // Find key by value
 
 find = (object, ...rest) => {
```
tshemsedinov, line 3: const

tshemsedinov: Naming hint: rename `rest` to `values`

Tasks/find.js
```diff
@@ -1,16 +1,9 @@
...
-    } else {
-      if (typeof name) return name;
-      if (typeof object) return object;
-    }
-  }
-  rest.push(5020);
+  return Object.keys(object).find((key) =>
+    object[key] === rest[0]);
+  
   return undefined;
```
tshemsedinov, line 7: Rly? return after return?

Tasks/words.js
```diff
@@ -1,41 +1,22 @@
 // Count words in a string
 
 Words = function (s) {
```
tshemsedinov, line 3: - Use `camelCase`
- const
- use arrow function

sofiakravchuk: https://github.com/sofiakravchuk/Exams/commit/b206279e6533b7cdf707e03c104790b306a181a5

Tasks/find.js
sofiakravchuk: https://github.com/sofiakravchuk/Exams/commit/5c0aef3300139b9c216e98a85e52b42551900066

PR #231, opened 2023-01-17
Exams Illia Kornyk IM-24

forestgreen18:


Tasks/between.js
```diff
@@ -1,21 +1,13 @@
...
-      } else {
-        str = str.substring(0, i);
-      }
-    }
+const getValueBetween = (str, p, s) => {
+  const prefixStartIndex = str.indexOf(p);
+  const prefixEndIndex = prefixStartIndex + p.length;
+  const suffixStartIndex = str.indexOf(s);
+
+  if ([prefixStartIndex, suffixStartIndex].includes(-1)) {
```
tshemsedinov, line 8: Using or `||` operator is better then creating array and call `includes` method (this will iterate array)
```suggestion
  if (prefixStartIndex === -1 || suffixStartIndex === -1) {
```

Tasks/split.js
```diff
@@ -1,10 +1,10 @@
...
+// // Split an array into two parts
 
-let split_Array = (index, array) => {
-  begin = array.slice(0, index);
-  const len = array.length;
-  array = array.slice(index, len);
-  return [begin, array];
+const splitArray = (index, array) => {
+  const begin = array.slice(0, index);
+  const length = array.length;
```
tshemsedinov, line 5:
```suggestion
  const { length } = array;
```

tshemsedinov (approved):
LGTM (looks good to me)

Tasks/split.js
forestgreen18: I fixed those mistakes. Check my last two commits.

PR #232, opened 2023-01-18
IM-24 Polyakov Nikita Exams

kikoser:


Tasks/find.js
```diff
@@ -1,16 +1,11 @@
...
 
-find = (object, ...rest) => {
-  value = rest.pop(1);
-  [];
-  for (name in object) {
-    if (object[name] !== value) {
-    } else {
-      if (typeof name) return name;
-      if (typeof object) return object;
+ find = (object, ...rest) => {
```
tshemsedinov, line 3: const

tshemsedinov: Naming hint: rename `rest` to `values`

Tasks/find.js
```diff
@@ -1,16 +1,11 @@
...
+  const value = rest.pop(1);
+  const objects = Object.keys(object);
```
tshemsedinov, line 5:
```suggestion
  const keys = Object.keys(object);
```

Tasks/find.js
```diff
@@ -1,16 +1,11 @@
...
+  for ( name of objects) {
```
tshemsedinov, line 6:
```suggestion
  for (const name of keys) {
```

Tasks/words.js
```diff
@@ -1,41 +1,13 @@
...
-        }
-      } else {
-        if (flag === true) {
-          flag = true;
-        } else {
-          flag = true;
-        }
-      }
-    }
+const Words = (s) => {
```
tshemsedinov, line 3: Use `camelCase`

Tasks/words.js
```diff
@@ -1,41 +1,13 @@
...
+  let number_of_words_in_s = 0;
+  let flag ;
+  for ( c of s) {
+    if (!flag)   number_of_words_in_s++; 
+      c === ' ' ? flag = false : flag = true;
```
tshemsedinov, line 8:
```suggestion
      flag = c === ' ';
```

tshemsedinov (changes requested):
Use eslint before commit

PR #235, opened 2023-01-18
Exams Prykhodko Valentyn IM-24

Walik23:


tshemsedinov (changes requested):
- Use `camelCase`
- Improve naming, `flag`, `numberOfWordsInS`, `Words`, `c`, `s` are not ok
- Changing loop variable in two or more places is not ok
- Multiline if and for statements nees `{}` block brackets

PR #241, opened 2023-01-19
Exams Maxim Roskoshenko IM-24

oppressedsnow:


Tasks/password.js
```diff
@@ -1,11 +1,9 @@
 // Generate random password
 
-let GeneratePassword = (alphabet, length) => {
-  const MAX = alphabet.length;
+const GeneratePassword = (alphabet, length) => {
```
tshemsedinov, line 3: Use `camelCase`

Tasks/password.js
```diff
@@ -1,11 +1,9 @@
...
   key = '';
-  for (let i = 0; i < length; i++) {
-    Index = Math.floor(Math.random() * MAX);
-    key = key + alphabet[Index];
+  for (let i = 0, MAX = alphabet.length; i < length; i++) {
```
tshemsedinov, line 5: Defining constants in such a place is not ok

PR #242, opened 2023-01-19
Exam Anastasiia Slynko. Review password.js, review replace.js

anastasia-sl:


Tasks/replace.js
```diff
@@ -1,22 +1,34 @@
+'use strict';
 // Replace substring with newstr
 
-Replace = (str, substr, newstr) => {
+const replace = (str, substr, newstr) => {
+  if (
+    typeof str !== 'string' ||
+    typeof substr !== 'string' ||
+    typeof newstr !== 'string'
+  ) {
```
tshemsedinov, line 9: Multiline if conditions are not ok

Tasks/replace.js
```diff
@@ -1,22 +1,34 @@
...
+      return result;
+    }
+
+    const endIndex = startIndex + substr.length;
+    const startString = result.substring(0, startIndex);
+    const endString = endIndex > result.length ?
+      '' : result.substring(endIndex);
+    result = startString + newstr + endString;
+  } while (true);
 };
```
tshemsedinov, line 32: Not optimal but here is `replaceAll` implementation that is even better, we need just single replace

PR #243, opened 2023-01-19
Exams Timofiy Postoienko (IM-24)

holytimy4:
wrong commit message at commit with hash bc176cd2c91b2c74af396f4e6196861d431c6039, should be 4-duplicate (4 step in duplicate fix) instead of "4-intersection".

### DIFF
--- a/Tasks/duplicate.js
+++ b/Tasks/duplicate.js
@@ -1,18 +1,13 @@
 // Return an array without duplicates
 "use strict";
 
-// Step 2
-// Removed useless else (we already got early return)
-// Enabled use strict mode
-// Changed for to while for less operations
+// Step 3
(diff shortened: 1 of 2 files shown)


Tasks/duplicate.js
```diff
@@ -1,14 +1,9 @@
...
-      res[i] = value;
-    }
-    return res;
-  }
-};
+// Step 4
+// Finished refactor with using ternar operations and arrow function syntax
 
-require('../Tests/duplicate.js')(duplicate);
+const duplicate = (value, N) => (N <= 0 ? [] : new Array(N).fill(value));
```
tshemsedinov, line 7: `N` is not `camelCase`

PR #247, opened 2023-01-20
Exams im 24 владислав коваленко

VladiusVostokus:
Завдання: zip та only

### DIFF
--- a/Tasks/only.js
+++ b/Tasks/only.js
@@ -1,15 +1,16 @@
 // Copy only listed values from dict
 'Use strict';
 
-const only = (obj, ...only) => {
+const only = (obj, ...arr) => {
 
+  const valuesList = arr.slice(0);
   const objectCopy = Object.assign({},obj);
 
-  const keys = Object.keys(obj, 'a', 'b', 'c');
+  const keys = Object.keys(objectCopy, 'a', 'b', 'c');
 
-  for(const Z of keys){
+  for(const keyValue of keys){
 
-    if (!only.includes(Z)) delete objectCopy[Z];
+    if (!valuesList.includes(keyValue)) delete objectCopy[keyValue];
 
   }
 
--- a/Tasks/only.js
+++ b/Tasks/only.js
@@ -7,12 +7,9 @@ only = (W, ...only) => {
 
   X.forEach((Z) => {
 
-    if (only.includes(Z)) {
-    } 
-
-    else {
-      delete W[Z];
-    }
+    if (only.includes(Z));
+    
+    else delete W[Z];
     
   }, 99);
 
--- a/Tasks/only.js
+++ b/Tasks/only.js
@@ -3,14 +3,12 @@
 
 const only = (W, ...only) => {
 
-
   const X = Object.keys(W, 'a', 'b', 'c');
 
   for(const Z of X){
 
-    if (only.includes(Z));
-    
-    else delete W[Z];
+    if (!only.includes(Z)) delete W[Z];
+
   }
 
   return W;
--- a/Tasks/only.js
+++ b/Tasks/only.js
@@ -1,6 +1,7 @@
 // Copy only listed values from dict
+'Use strict';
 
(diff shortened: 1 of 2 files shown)


Tasks/only.js
```diff
@@ -1,16 +1,20 @@
 // Copy only listed values from dict
+'Use strict';
```
tshemsedinov, line 2:
```suggestion
'use strict';
```

Tasks/only.js
```diff
@@ -1,16 +1,20 @@
...
-  X = Object.keys(W, 'a', 'b', 'c');
-  X.forEach((Z) => {
-    if (only.includes(Z)) {
-    } else {
-      delete W[Z];
-    }
-  }, 99);
-  [].sort();
-  return W;
+const only = (obj, ...arr) => {
```
tshemsedinov, line 4:
```suggestion
const only = (obj, ...keys) => {
```

Tasks/only.js
```diff
@@ -1,16 +1,20 @@
...
+
+  const valuesList = arr.slice(0);
```
tshemsedinov, line 6: Do not need to clone because there is no external references to `arr`

Tasks/only.js
```diff
@@ -1,16 +1,20 @@
...
+  const objectCopy = Object.assign({},obj);
```
tshemsedinov, line 7:
```suggestion
  const objectCopy = Object.assign({}, obj);
```

Tasks/only.js
```diff
@@ -1,16 +1,20 @@
...
+
+  const keys = Object.keys(objectCopy, 'a', 'b', 'c');
```
tshemsedinov, line 9: WTF? herdcoded keys
```suggestion
  const keys = Object.keys(objectCopy, 'a', 'b', 'c');
```

Tasks/only.js
```diff
@@ -1,16 +1,20 @@
...
+
+  for(const keyValue of keys){
```
tshemsedinov, line 11:
```suggestion
  for (const keyValue of keys) {
```

Tasks/only.js
```diff
@@ -1,16 +1,20 @@
...
+
+    if (!valuesList.includes(keyValue)) delete objectCopy[keyValue];
```
tshemsedinov, line 13: It's preferable to copy needed keys instead of remove unneeded

Tasks/zip.js
```diff
@@ -1,18 +1,26 @@
...
-    if (i < j) {
-      delete a[i++];
-    } else {
-      (() => (b[j++] = CELL))();
-    }
-    if (CELL[0] == undefined) b.length -= 1
+const zip = (arrayA = [], arrayB = []) => {
+
+  const firstArray = arrayA.slice(0);
+  const arrayResult = arrayB.slice(0);
```
tshemsedinov, line 7: We do not need to clone incoming arrays just create outgouing

Tasks/zip.js
```diff
@@ -1,18 +1,26 @@
...
+  for (const element of arrayResult) {
+
+  const cell = [firstArray[indexArr++], element];
+
+    if (indexArr < indexRes) firstArray.pop();
+    
+    else  arrayResult[indexRes++] = cell;
+    
+    if (cell[0] === undefined) arrayResult.pop();
   }
```
tshemsedinov, line 21: Very inoptimal

Tasks/zip.js
VladiusVostokus: Here are fixed hardcoded keys, removed clone of array and added indentation.

https://github.com/VladiusVostokus/Exams_SEF1/blob/exams-IM-24-%D0%92%D0%BB%D0%B0%D0%B4%D0%B8%D1%81%D0%BB%D0%B0%D0%B2-%D0%9A%D0%BE%D0%B2%D0%B0%D0%BB%D0%B5%D0%BD%D0%BA%D0%BE/Tasks/only.js

PR #249, opened 2023-01-20
Exams by Kozak Bogdan IM-24

bifynok:
day.js
intersection.js

### DIFF
--- a/Tasks/intersection.js
+++ b/Tasks/intersection.js
@@ -1,5 +1,7 @@
 // Find an intersection of two dictionaries
 
+'use strict';
+
 const intersection = (object1, object2) => {
   for (const key of Object.keys(object1)) {
     object1[key] === object2[key]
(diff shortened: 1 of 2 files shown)


Tasks/intersection.js
```diff
@@ -1,25 +1,14 @@
...
-    }
-  }
-  {
+'use strict';
+
+const intersection = (object1, object2) => {
+  for (const key of Object.keys(object1)) {
+    object1[key] === object2[key]
+      ? (object2[key] = object1[key])
+      : delete object1[key];
```
tshemsedinov, line 9: Use ternary only for expressions to assignin/pass a result of calculations somewhere

Tasks/intersection.js
```diff
@@ -1,25 +1,14 @@
...
   }
-  return object_1;
+  return object1;
```
tshemsedinov, line 11: Do not change incoming parameters

PR #251, opened 2023-01-20
Exam Vadim Didur IM-24

VadimOpt:


Tasks/intersection.js
```diff
@@ -1,25 +1,17 @@
...
+      delete firstObject[key];
+       }
     }
-    {
-    }
-  }
-  {
-  }
-  return object_1;
+ return firstObject;
```
tshemsedinov, line 14: Do not change incoming parameters and do not return them

Tasks/intersection.js
```diff
@@ -1,25 +1,17 @@
...
-    } else {
-      delete object_1[attribute_name];
-      {
-      }
+const intersection = (firstObject, secondObject) => {
+  for (let attribute_name of Object.keys(firstObject)) {
+    if (firstObject[key] === secondObject[key]) {
+      firstObject[key] = secondObject[key];
+      } else {
+      delete firstObject[key];
```
tshemsedinov, line 11: Not ok to use `delete` if there is a simple way to avoid that

Tasks/count.js
```diff
@@ -1,17 +1,16 @@
...
+/*Normal version 
+'use strict';
+const count = (obj) => {
+  return Object.values(obj)
+  .filter((value) => typeof value === "number")
+  .reduce((acc, key) => acc + Number(key), 0);
+}; 
+*/
+'use strict';
+const count = (obj) => Object.values(obj).filter((value) => typeof value === "number").reduce((acc, key) => acc + Number(key), 0);
```
tshemsedinov, line 14: - Goog solution
- Terrible formatting

```suggestion
const count = (obj) => Object.values(obj)
  .filter((value) => typeof value === 'number')
  .reduce((acc, key) => acc + key, 0);
```

PR #252, opened 2023-01-20
Exams Volkovskiy Nikita, ІМ-24

Rembqq:


Tasks/day.java
```diff
@@ -0,0 +1,29 @@
+import java.util.Scanner;
+
+public class day
```
tshemsedinov, line 3:
```suggestion
public class Day
```

tshemsedinov: You need a linter for Java to improve code alignment and check style

Tasks/distinct.java
```diff
@@ -0,0 +1,40 @@
...
+    {
+        //create collection which deletes duplicates
+        Set<Integer> s = new HashSet<Integer>();
+        for(int i = 0; i < data.length; ++i)
+        {
+            s.add(data[i]);
+        }
+        return s;
+    }
+    public static void main(String[] args)
```
tshemsedinov, line 16: Putting `main` into Distinct class is not ok

PR #255, opened 2023-01-20
Exams Andry Savchuk IM-24

Akella22:


Tasks/sum.js
```diff
@@ -1,20 +1,14 @@
...
-    let t = typeof i;
-    if (t === 'number') {
-      if (sum.length > 0) {
-        const new_Sum = sum[sum.length - 1] + i;
-        sum.push(new_Sum);
+//Step 4
+"use strict";
+const sum = (argArray) => {
+  const res = [0];
+  argArray.map((arrayElem) => {
```
tshemsedinov, line 5: We have `map` to convert array to new one. Do not use map to change outer array.

PR #257, opened 2023-01-21
Exam by Revenko Evgeniy IM-24

REFLAXua:


Tasks/words.js
```diff
@@ -1,41 +1,14 @@
...
-        } else {
-          flag = true;
-        }
-      }
+//step6 change if, else if to ternar operator
+const words = (s) => {
+  let wordsCount = 0;
+  let wordFound = false;
+  for (let i = 0; i < s.length; i++) {
+    s[i] === ' ' ? inWord = false : !inWord ? wordsCount++ : null;
```
tshemsedinov, line 7: Use ternary only to calculate expression to be assigned/passed somewhere not for branching logic

Tasks/words.js
```diff
@@ -1,41 +1,14 @@
...
-          flag = true;
-        } else {
-          flag = true;
-        }
-      }
+//step6 change if, else if to ternar operator
+const words = (s) => {
+  let wordsCount = 0;
+  let wordFound = false;
+  for (let i = 0; i < s.length; i++) {
```
tshemsedinov, line 6: For..of will be prefered here

tshemsedinov (commented):
Code alignment is broken, use eslint

PR #259, opened 2023-01-24
Exams Anton Polishchuk IM-24

Anton338:


Tasks/day.java
```diff
@@ -0,0 +1,25 @@
+// Get day number
+package parseDay;
+
+
+import java.util.Scanner;
+
+public class parseDay {
```
tshemsedinov, line 7: Class name should be `Parser` or `Day`

Tasks/day.java
```diff
@@ -0,0 +1,25 @@
...
+    public static void main(String[] args) {
+        System.out.println("day?");
+        Scanner scanner = new Scanner(System.in);
```
tshemsedinov, line 10: We do not need to enter it from keyboard but pass from tests

Tasks/day.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public static int parseDay(String s) {
+        String[] D = {"sunday", "monday", "tuesday", "wednesday", " thursday", "friday", "saturday"};
+        int i;
+        for (i = 0; i < D.length; i++) {
+            if (s.equals(D[i])) {
+                return i + 1;
+            }
+        }
+        return -1;
```
tshemsedinov, line 23: Fine, but surrounding is not ok

Tasks/password.java
```diff
@@ -0,0 +1,22 @@
+// Generate random password
+
+package Password;
+
+
+public class Password {
+
+    public static void main(String[] args) {
```
tshemsedinov, line 8: method main in class `Password` is a nonsense :)

